PR #27, opened 2014-12-22
don't always set ace/mode/coffee

cpirich:
ace editor was not being set to ace/mode/javascript when droplet was initialized in javascript mode.

please verify that this works properly in mode/coffee as I don't have a good way to verify for sure.

also, I don't know if this will also handle dynamic calls to droplet's setMode.


davidbau:
Tested, and it still works fine with coffeescript. Merging.

PR #28, opened 2015-02-18
Add support for dotted methods in Javascript and Coffeescript

davidbau:
Dotted methods can now be listed in function whitelists; if a fully-qualified function name such as "Math.round" matches a dotted name, the function name is treated as a single (non-editable) block.


cpirich:
Looks good!

PR #29, opened 2015-02-23
Coffeescript: merge inline code definition blocks.

davidbau:
This change (requested by CS First instructors)
avoids treating inline lambda (->) functions as
separate blocks in Coffeescript in two cases:

(1) direct assignment of a lambda

assignable = (x) -> dosomething

Instead of treating the = and the -> as separate blocks, these are
treated as a single block.

(2) use of a lambda as the last argument of a function call

button 'clickme', ->
  dosomething

Instead of treating the function call (button(..)) as a separate
block from the ->, these are treated as a single block.

This is implemented by factoring the handling of a -> code block
into an @addCode method, and separating the socket handling from
the delineation of the block for the ->.  @addCode is called
by the Assign handler when the RHS is a Code node and by the
Call handler with the last arg is a Code node.

Other Code nodes still get their own block, so, for example,
putting the code in parentheses will split it into its own block.

### DIFF
--- a/src/coffee.coffee
+++ b/src/coffee.coffee
@@ -322,9 +322,18 @@ define ['droplet-helper', 'droplet-model', 'droplet-parser', 'coffee-script'], (
       if last? and last.value in list then return 1
       return 0
 
+    # ## addCode ##
+    # This shared logic handles the sockets for the Code function
+    # definitions, even when merged into a parent block.
+    addCode: (node, depth, indentDepth) ->
+      for param in node.params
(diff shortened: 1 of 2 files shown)


caitsydney:
Woot! :) Thanks for the quick turnaround. Can't wait to see it live!

cpirich:
Any chance that we could use this for JavaScript as well (perhaps optionally?)

If so, it would be nice if it didn't require the function to be on the last parameter. The setTimeout() function has a function as the first parameter and the time value as the second parameter.

PR #30, opened 2015-02-23
Support UnaryExpression nodes like !() in Javascript.

davidbau:
This change adds support for (value-colored) blocks that come from
Javascript UnaryExpressions such as ~() and !().  Previously these
nodes were unhandled, so would just be greated as unblockified
plain text areas.

### DIFF
--- a/src/javascript.coffee
+++ b/src/javascript.coffee
@@ -116,6 +116,7 @@ define ['droplet-helper', 'droplet-model', 'droplet-parser', 'acorn'], (helper, 
 
   COLORS = {
     'BinaryExpression': 'value'
+    'UnaryExpression': 'value'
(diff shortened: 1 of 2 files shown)


cpirich:
Looks good. I don't understand the meaning of the precedence value 17, but otherwise it all seems to make sense. I found this table online, but it seems to suggest that 15 is the right value: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Operator_Precedence

davidbau:
Good find, thanks for the catch.  Actually, the numbers we use for precedence increase in the opposite sense (lowest is tightest).  I've brought all our numbers in alignment with that table, using (19-x).

PR #34, opened 2015-03-06
Csv adapter with self parser

coolyashish:


davidbau:
This is an improved version, right?  I'll close out the other pull request with the previous version.

coolyashish:
Updated repository here : https://bitbucket.org/run_009/droplet

PR #35, opened 2015-03-09
Dropdowns

dabbler0:
Add support for dropdown sockets. Sockets with dropdowns have a black arrow next to them; the menu is brought up when the arrow is clicked. Otherwise, the socket can be typed in normally.

Specified through modeOptions by adding a "dropdown" property to a function description object, like

```
"pen": {
  dropdown: {
    0: [
      {text: 'red', display: '<span style="color:red">red</span>'}
      {text: 'blue', display: '<span style="color:blue">blue</span>'}
    ]
  }
}
```

This `display` property can be arbitrary HTML.


davidbau:
The description of the config in the PR isn't accurate any more: you now map from position to a list of strings (maybe it's an option to map to objects with HTML, but it's not tested).

Can you
- Add a quick test that exercises a non-zero arg index, and an HTML menu.
- Add documentation to embedding.md describing how to do it.

davidbau:
Looks great!

PR #36, opened 2015-03-10
Add more colors.

davidbau:
We now support #hex colors as well as a set of color names
for nicely deistinc colors such as "amber" or "lightblue" taken
from the 200 row of http://www.materialui.co/colors (Which is Google's
"Material Design" recommended colors.)


dabbler0:
Looks good to me.

PR #39, opened 2015-03-12
Contexts

dabbler0:
Add support for parsing contexts. A parsing context is a prefix, suffix, and indentation level that are added to a block before it reparses, then removed afterward. For instance, the coffeescript `when` block:
```
when a
  do b
```
Has the parsing context with prefix `switch\n`, no suffix, and an indent of `  `. So before being passed to the parser for reparsing, it becomes:
```
switch
  when a
    do b
```
Then after parsing, the parser wrapper layer looks for the block at the correct location in the result, clips it out, and returns that, to give a block that reads with the original text..


src/coffee.coffee
```diff
@@ -916,15 +936,15 @@ define ['droplet-helper', 'droplet-model', 'droplet-parser', 'coffee-script'], (
     if n < 0 or n >= lines.length
       return false
     # Refuse to add another empty backtick line if there is one already
```
davidbau: s/backtick/underscore/

src/controller.coffee
```diff
@@ -1252,9 +1256,9 @@ define ['droplet-helper',
         else
           newBlock.rawReplace oldBlock
 
-    catch e
-      throw e
-      return false
+    #catch e
```
davidbau: why removed?

src/model.coffee
```diff
@@ -526,6 +526,8 @@ define ['droplet-helper'], (helper) ->
 
       @version = 0
 
+    getStringRepresentation: -> ''
```
davidbau: add a comment describing the role and use of this method in the interface.

davidbau:
It's odd that this merge would leave us in a state when switch cases can be dragged around, but then they can be dragged out to regular statement context where they're illegal.

What would it take to:

(1) Configure an enumeration of parsing contexts on the side and
(2) Tackle droppability rules at the same time?

PR #41, opened 2015-03-15
Fix palette highlight position and empty header.

davidbau:
When the palette is resized (after changing the palette header),
the palette highlight canvas should also be resized, so the
highlight outlines are positioned correctly.

Also, if the palette header has only a single group and it
has no name, then the palette header is formatted so that
the palette header is not visible.


dabbler0:
Looks good to me.

PR #42, opened 2015-03-19
option to always show palette

Bjvanminnen:
keeppalette

This adds an option to keep the palette around in code mode. Not sure whether this better belongs as a child of modeOptions?

I also added a couple of tests, and verified that when alwaysShowPalette is false or undefined, the behavior remains unchanged.

Haven't really written coffee script before, so let me know if there are coffee script conventions I've violated.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -1696,7 +1696,10 @@ define ['droplet-helper',
           @redrawTextInput()
 
   Editor::resizeAceElement = ->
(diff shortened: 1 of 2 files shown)


Bjvanminnen:
Closing this PR as all the necessary code will also be in Chris's PR.

cpirich:
These changes are now included in https://github.com/dabbler0/droplet/pull/43

PR #43, opened 2015-04-01
Options for willShowPaletteWithText and explicit showPalette method

cpirich:
This PR combines the work from @bjvanminnen in https://github.com/dabbler0/droplet/pull/42 with additional work to explicitly call a showPalette() method to hide or show the palette.

* New option: enablePaletteAtStart - allows the host to explicit when the Editor is first created whether or not the palette should be visible
* Renamed option: Brent added alwaysShowPalette - I've renamed that to be showPaletteInTextMode since that option now determines whether or not the palette is allowed to appear when in ace/text mode
* New method: enablePalette(enabled) - This can be called with 'true' to make the palette appear or disappear
* New property: paletteEnabled - this determines whether or not the palette should appear (NOTE that it is ignored if in text mode when showPaletteInTextMode is false - this preserves the existing behavior for hosts that don't use these new options)
* The palette animations from melt/freeze are extracted into enablePalette. They will continue to run as part of melt/freeze if necessary (switching block/text mode with paletteEnabled == true and showPaletteInTextMode == false)
* I added one additional unit test and updated the two unit tests that Brent created

NEW Bonus: I fixed the droplet canvas disappearing problem. It was caused by the hiddenInput being absolutely positioned outside of the client area of the dropletElement. When it start handling text input, the browser scrolled to it, sliding all of the other sibling divs far away from their intended positions.


src/controller.coffee
```diff
@@ -140,6 +140,8 @@ define ['droplet-helper',
   exports.Editor = class Editor
     constructor: (@wrapperElement, @options) ->
       @paletteGroups = @options.palette
+      @willShowPaletteWithText = @options.willShowPaletteWithText ? false
```
Bjvanminnen: I found my brain having trouble parsing this option name (i.e. I was reading as will there be text in the palette instead of will there be a palette in tell mode). What about willShowPaletteInTextMode or showPaletteInTextMode?

Bjvanminnen: Another approach might be to call it something like togglePaletteOnTransition which defaults to true.

src/controller.coffee
```diff
@@ -1695,7 +1701,11 @@ define ['droplet-helper',
           @redrawTextInput()
 
   Editor::resizeAceElement = ->
-    @aceElement.style.width = "#{@wrapperElement.offsetWidth}px"
+    width = @wrapperElement.offsetWidth
+    if @willShowPaletteWithText and @showingPalette
```
Bjvanminnen: "New property: showingPalette - this determines whether or not the palette should appear (NOTE that it is ignored if in text mode when willShowPaletteWithText is false - this preserves the existing behavior for hosts that don't use these new options)"

This seems confusing that showingPalette doesn't necessarily reflect if we're showing the palette (i.e. if willShowPaletteWithText is false). I would expect showingPalette to always be accurate, and our other setting to determine whether the state of showingPalette changes on melt/freeze.

src/controller.coffee
```diff
@@ -3319,6 +3342,60 @@ define ['droplet-helper',
...
+          @paletteWrapper.style.top = '-9999px'
+          @paletteWrapper.style.left = '-9999px'
+
+          @currentlyAnimating = false
+        ), 500
+
+      else
+        @paletteWrapper.style.top = '0px'
+        @paletteWrapper.style.left = "#{-@paletteWrapper.offsetWidth}px"
+        @paletteHeader.style.zIndex = 257
```
Bjvanminnen: Why 257? This number seems pretty arbitrary.

src/controller.coffee
```diff
@@ -3118,28 +3136,38 @@ define ['droplet-helper',
...
         # Translate the ICE editor div out of frame.
         @dropletElement.style.transition =
           @paletteWrapper.style.transition = ''
 
-        @dropletElement.style.top = '-9999px'
-        @dropletElement.style.left = '-9999px'
+        # Translate the ACE editor div into frame.
+        @aceElement.style.top = '0px'
+        if @showPaletteInTextMode and @paletteEnabled
+          @aceElement.style.left = @paletteWrapper.style.width
```
davidbau: Works better if we position the aceElement based on paletteWrapper.offfsetWidth.  (In case there's some CSS such as maxWidth that overrides the effective width of the paletteWrapper, as happens in pencil code).  Tweaking this on merge.

davidbau:
Merged in e4d7edb003c823d349b27dee04da7cf3119735e2.

PR #46, opened 2015-04-09
Pull request for CSV adapter.

sakagg:
I have initiated a pull request so that it becomes easier for you to review changes made by me. I also removed the extra files and synced with upstream before initiating the PR.


.gitignore
```diff
@@ -7,3 +7,4 @@ example/*.js
 example/*.map
 test/*.js
 test/*.map
+TEST_2G
```
davidbau: What's test_2g?

src/controller.coffee
```diff
@@ -1853,13 +1878,13 @@ define ['droplet-helper',
       # value.
       unless @textFocus.handwritten
         newParse = null
-        string = @textFocus.stringify(@mode.empty).trim()
+        string = @mode.normalString @textFocus.stringify(@mode.empty)
```
davidbau: might we want all the logic from line 1881 to 1890 to be part of the base Parser normalString, i.e., trim and look at quotes?  Then specific modes can override to have their whole own idea of what quotes are etc.

src/controller.coffee
```diff
@@ -4003,18 +4035,17 @@ define ['droplet-helper',
 
     if line of @lineNumberTags
       lineDiv = @lineNumberTags[line]
-
+
```
davidbau: search and remove trailing spaces on lines of your code.

src/controller.coffee
```diff
@@ -958,6 +961,28 @@ define ['droplet-helper',
...
+    mainPoint = @trackerPointToMain point
+    hitTestResult = @hitTest mainPoint, @tree
+
+    if hitTestResult?
+      hitTestBlock = @view.getViewNodeFor hitTestResult
+      lineNumber = @findLineNumberAtCoordinate(mainPoint.y)
+
+      if hitTestBlock.addButtonRect? and hitTestBlock.addButtonRect.contains mainPoint
+        console.log "Plus on: ", lineNumber
```
davidbau: remove console.log

src/view.coffee
```diff
@@ -1642,12 +1646,38 @@ define ['droplet-helper', 'droplet-draw', 'droplet-model'], (helper, draw, model
...
+        super
+
+        start = @totalBounds.x + @totalBounds.width - @extraWidth
+
+        ctx.textBaseline = 'top'
+        ctx.font = @view.opts.textHeight + 'px '
+        ctx.fillStyle = '#000'
+
+        if 'add-button' in @model.classes
+          @addButtonRect.stroke(ctx, '#000')
```
davidbau: instead of stroking the rect directly, try getting a path (via rect.toPath) and then using the path's beveling feature to draw a beveled path.  Can you try it to see how it looks?

.gitignore
sakagg: That's just a folder of for some parsers I was trying. It contains various CSS and HTML parsers I was looking at while making my proposal. I'll remove that in the next commit.

sakagg:
Hey
 - I've fixed the above mentioned issues.
 - Added tests for CSV Adapter, but I am still looking for a way t test button.
 - As far as bevels are concerned, I am getting strange errors on setting the bevel variable to true so I have removed that part for now. I'll fix them and update the repo.

src/controller.coffee
```diff
@@ -1948,6 +1968,15 @@ define ['droplet-helper',
 
     socket.notifyChange()
 
+  Editor::populateBlock = (block, string) ->
+    newBlock = @mode.parse(string, wrapAtRoot: false).start.next.container
+
+    if newBlock?.start?.next?
+      block.start.append newBlock.start.next
+      newBlock.end.prev.append block.end
+      block.notifyChange()
```
davidbau: It's interesting that you replace the contents of the block but do not replace the block itself, which is counterintuitive.  Was there a reason that you needed to do that?  Also, I'd love Anthony to weigh in on whether anything is missing here.

src/controller.coffee
```diff
@@ -958,6 +961,28 @@ define ['droplet-helper',
...
+    mainPoint = @trackerPointToMain point
+    hitTestResult = @hitTest mainPoint, @tree
+
+    if hitTestResult?
+      hitTestBlock = @view.getViewNodeFor hitTestResult
+      lineNumber = @findLineNumberAtCoordinate(mainPoint.y)
+
+      if hitTestBlock.addButtonRect? and hitTestBlock.addButtonRect.contains mainPoint
+        #console.log "Plus on: ", lineNumber
+        line = @mode.handleButton @getValue().split('\n')[lineNumber], 'add-button'
```
davidbau: handleButton should not work at the line-level, but at the block level.

You should probably pass handleButton three arguments
(1) the entire text for the block (which may be multiline)
(2) possibly the set of classes for the block (so that it's easy to tell which type of block it is)
(3) the command - 'add-button'

David

src/controller.coffee
sakagg: I did that intentionally. This is because if the users removes everything from the block, the block itself should not disappear, only the contents should. In other words I can say that the buttons should have the capability to modify the contents of the block, not the block itself (or it's structure).
Here is a pictorial representation.
Original Block
original
If contents are replaced
replace_contents
If entire block is replaced
replace_block


On the other hand, we can have cases when the block should also be modified (eg: hide the - button if no children left in block) in which case we should replace the entire block. This will require modifying the parser such that is parses an empty string to something meaningful.

src/controller.coffee
sakagg: I felt the same and suggested it few days back on the google group (https://groups.google.com/forum/#!msg/pencilcode-development/kcdiFv6GtdA/mTYMcqyX9EUJ), but since I didn't get an affirmation, I didn't work on it. Now that you have confirmed it, I went ahead and implemented it and you can see it in the latest commit.

Saksham

src/view.coffee
```diff
@@ -1642,12 +1646,46 @@ define ['droplet-helper', 'droplet-draw', 'droplet-model'], (helper, draw, model
...
+        if 'add-button' in @model.classes
+          @extraWidth += @view.opts.buttonWidth + @view.opts.buttonPadding
+
+        if 'subtract-button' in @model.classes
+          @extraWidth += @view.opts.buttonWidth + @view.opts.buttonPadding
+
         for size, i in @minDimensions
           size.width = Math.max size.width,
               @view.opts.tabWidth + @view.opts.tabOffset
+          size.width += @extraWidth
```
davidbau: For a multiline block, you're adding @extrawidth to every line.  The +/- button should be on the last line, so should just apply there.

src/view.coffee
```diff
@@ -1642,12 +1646,46 @@ define ['droplet-helper', 'droplet-draw', 'droplet-model'], (helper, draw, model
...
 
         return null
 
+      drawSelf: (ctx, style) ->
+        super
+
+        start = @totalBounds.x + @totalBounds.width - @extraWidth
```
davidbau: Need to think through positioning in the multiline case.  Currently you put the +/- buttons in the upper-right corner of the bounding rect for the block.  We want it to be on the last line.  But lower-right corner isn't right: you should figure the actual width of the last line to position the +/-.

PR #47, opened 2015-04-10
Support empty indents in Javascript

davidbau:
When a Javascript block is empty but surrounded by curly braces, it should be serialized with nothing between the curly braces.  This is different from coffeescript, which needs something inside the indented code.

This change adds support for allowing a language adapter to support "empty indents".


davidbau:
This solves the problem, in javascript, of empty curly braces.  Instead of inserting a "__" inside empty curlies when serializing, we just leave them empty.

PR #48, opened 2015-04-24
Block to text

Bjvanminnen:
This allows us to drag blocks into the editor when in text mode and have the text inserted.

dragtext


src/controller.coffee
```diff
@@ -834,33 +833,31 @@ define ['droplet-helper',
...
+      container = container.visParent()
 
-    node.setLeadingText leading; node.setTrailingText trailing
+    @prepareNode node, container
+    node.spliceIn location
 
-    node.spliceOut()
 
-  Editor::spliceIn = (node, location) ->
+  Editor::prepareNode = (node, context) ->
```
Bjvanminnen: Not sure what a better name for this would be. I noticed there was shared logic between spliceIn and spliceOut, and this was logic that I also wanted to call without a context when dropping in ace mode (so that I get the benefit of the call to parens)

src/javascript.coffee
```diff
@@ -515,7 +515,9 @@ define ['droplet-helper', 'droplet-model', 'droplet-parser', 'acorn'], (helper,
        node.type is 'segment'
       trailing trailing().replace(/;?\s*$/, '')
     else
-      trailing trailing().replace(/;?\s*$/, ';')
```
Bjvanminnen: The function that this is in (parens) already seems like it should/could be named better, considering adding semicolons doesn't really have to do with parens. Not sure what a good name is that covers both what happens in JS land and in coffee land.

cpirich:
LGTM

src/controller.coffee
```diff
@@ -1179,65 +1184,95 @@ define ['droplet-helper',
...
+
+          # Call prepareNode, which may append with a semicolon
+          @prepareNode @draggingBlock, null
+          text = prefix + @draggingBlock.stringify @mode
+
+          if not prefix and text[text.length - 1] == ';'
+            # Add a potentially indented new line if
+            # we didn't already do so.
+            text += '\n' + leadingWhitespace
+          @aceEditor.onTextInput text
```
Bjvanminnen: Not sure if all of this logic around adding newlines/indentation makes more sense existing elsewhere.

cpirich:
Closing this. It is now incorporated into https://github.com/dabbler0/droplet/pull/50

PR #52, opened 2015-05-12
Handle indentation for multi-line blocks

Bjvanminnen:


Bjvanminnen:
I still don't have write access, so I can't merge this myself.

Bjvanminnen:
@cpirich If these changes look okay to you, I'll also update lib/droplet in the cdo repo.

src/controller.coffee
```diff
@@ -1202,11 +1202,18 @@ define ['droplet-helper',
             # We're at the end of a non-empty line.
```
cpirich: This comment may need to be updated since it doesn't seem like the indentation is handled in this clause anymore. It isn't clear to me from reading the new code how leadingWhiteSpace is added to the first line (since it is skipped if index==0 down below)

Bjvanminnen:
Closing PR as this change isn't quite right.

PR #53, opened 2015-05-14
Handle indentation for multi-line blocks

Bjvanminnen:
Another attempt at multiline indentation, this one better.

The only case that's a little bit weird still is if I insert at the end of a line and had removed the indentation from the following line (see the last thing I did in the attached screencap)

indentation


cpirich:
LGTM

PR #54, opened 2015-05-20
Add expansion blocks to palette

davidbau:
This feature allows a palette block to "expand" into another block when
dragging out of the palette.  The idea is that the block rendered in
the palette can be an abbreviation, and the "real" block might be bigger,
for example, it might include longer arguments etc.

To configure an expansion block, do the following in the palette configuration:

```js
      palette: [{
        name: 'Draw',
        color: 'blue',
        blocks: [{
          block: 'pen()',       // this is what you see in the palette
          expansion: 'pen red', // this is what you get when you drag
          title: 'ptest'
        }],
```

This commit also adds a basic uitest that simulates mouse events
to test the controller, which exercises the expansion feature.

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -7,6 +7,15 @@ notify = (message) ->
     fallback: 0
   ), ->
 
+serveNoDottedFiles = (connect, options, middlewares) ->
+  # Avoid leaking .git/.svn or other dotted files from test servers.
+  middlewares.unshift (req, res, next) ->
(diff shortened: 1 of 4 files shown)


cpirich:
Looking forward to trying this out! We had discussed supporting expansion as a callback function as well. Is that still an option? We would use it to generate unique parameter names.

davidbau:
Ah, forgot - easy enough to add.  I'll add it.

David

PR #55, opened 2015-05-27
add readOnly, textModeAtStart, fix enablePaletteAtStart & setEditorState

cpirich:
* `getReadOnly` and `setReadOnly` implemented, matching the Ace API (and calling down into Ace for `setReadOnly`)
* `@readOnly` is checked on the droplet side before hit testing, input focus, and key events. The droplet editor can still be scrolled, lines can be highlighted, but no modifications to the content of the document should be possible
* `setEditorState` didn't properly check `@paletteEnabled` and `@showPaletteInTextMode` to position the palette off-screen when necessary. That code only existed in the async `performMeltAnimation` and `performFreezeAnimation`. It now works either way.
* `enablePaletteAtStart` mostly worked if `false` was passed in, but there was a bug in that the palette was never moved off-screen. Solved by always calling `setEditorState` in the constructor.
* `textModeAtStart` is a new option that can be passed to the constructor if you wish to start in text mode

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -3854,7 +3854,8 @@ define ['droplet-helper',
 
   Editor::setEditorState = (useBlocks) ->
     if useBlocks
-      @setValue @getAceValue()
+      unless @currentlyUsingBlocks
+        @setValue @getAceValue()
 
       @dropletElement.style.top = '0px'
       if @paletteEnabled
@@ -3873,13 +3874,15 @@ define ['droplet-helper',
         @highlightCanvas.opacity = 1
 
       @resizeBlockMode(); @redrawMain()
-
+      
     else
       paletteVisibleInNewState = @paletteEnabled and @showPaletteInTextMode
 
       oldScrollTop = @aceEditor.session.getScrollTop()
 
-      @setAceValue @getValue()
+      if @currentlyUsingBlocks
+        @setAceValue @getValue()
+
       @aceEditor.resize true
 
       @aceEditor.session.setScrollTop oldScrollTop
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -305,9 +305,9 @@ define ['droplet-helper',
 
       # If we were given an unrecognized mode or asked to start in text mode,
       # flip into text mode here
-      blockMode = @mode? && !@options.textModeAtStart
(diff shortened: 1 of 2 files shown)


src/controller.coffee
```diff
@@ -302,9 +303,11 @@ define ['droplet-helper',
...
       @rebuildPalette()
 
-      # If we were given an unrecognized mode, flip into text mode
-      unless @mode?
-        @setEditorState false
+      # If we were given an unrecognized mode or asked to start in text mode,
+      # flip into text mode here
+      blockMode = @mode? && !@options.textModeAtStart
+      # Always call @setEditorState to ensure palette is positioned properly
+      @setEditorState blockMode
```
Bjvanminnen: I might call this inBlockMode or useBlockMode to make it a little bit more clear that this is a bool, unlike @mode

src/controller.coffee
```diff
@@ -645,6 +657,9 @@ define ['droplet-helper',
   # Simple function for going through a linked-list block
   # and seeing what the innermost child is that we hit.
   Editor::hitTest = (point, block) ->
+    if @readOnly
+      return null
+
```
Bjvanminnen: For this and trackerPointIsInElement, it's not clear to me in isolation why these functions should exiting early in readOnly mode. I'm not entirely sure what a trackerPoint is, but I would expect that it is/isn't in an element independent of whether or not we're in read only mode. For hit test, my confusion might come from the fact that it's not entirely clear to me what it means to find a hit.

Bjvanminnen:
Couple nits, but otherwise lgtm

src/controller.coffee
cpirich: hit testing (in this case) is checking to see if mouse coordinates are considered a "hit" against an "element" or "block" in the UI. By failing all hit tests on droplet's inner elements, it make it impossible to select, drag, set focus, etc.

src/controller.coffee
cpirich: Good suggestion. Will fix.

davidbau:
Integration tests on travis seem to be failing for these changes (test/test.html times out).

And it repros on my local "grunt test", although test/test.html does pass in-browser in Chrome.  Can you track down which part o test.html is timing out?

PR #62, opened 2015-06-08
Allow dropdown options to specify their own `click` handler

joshlory:
In Applab we'd like to be able to have dropdowns with an "Upload new..." option that opens a modal dialog.  This PR lets dropdown options specify a click handler to be called instead of the default handler.

screen shot 2015-06-08 at 3 48 18 pm

Sample usage:
```javascript
{display: 'Upload new...', click: (callback) -> promptImage(callback)}
```

The provided handler receives a callback which it can call to populate the socket value (usually after another user interaction).

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -2179,14 +2179,21 @@ define ['droplet-helper',
         div.style.fontSize = @fontSize
         div.style.paddingLeft = helper.DROPDOWN_ARROW_WIDTH
 
-        handleClick = (text) =>
+        setText = (text) =>
+          # Attempting to populate the socket after the dropdown has closed should no-op
+          return if @dropdownElement.style.display == 'none'
+
           @populateSocket @textFocus, text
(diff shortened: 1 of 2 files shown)


src/controller.coffee
```diff
@@ -2179,12 +2179,14 @@ define ['droplet-helper',
...
-          @populateSocket @textFocus, el.text
-          @hiddenInput.value = el.text
+        handleClick = (text) =>
+          @populateSocket @textFocus, text
+          @hiddenInput.value = text
 
           @redrawMain()
           @hideDropdown()
+
+        div.addEventListener 'mouseup', el.click?.bind(null, handleClick) or handleClick.bind(null, el.text)
```
Bjvanminnen:
```
setText(text) => 
  @populateSocket @textFocus, text
  ...

div.addEventListener 'mouseup', =>
  if el.click
    el.click(setText)
  else
   setText(el.text)

PR #64, opened 2015-06-10
Add bubbling for reparses, and improve error handling for reparsing.

dabbler0:
Incorporates four changes to reparsing behavior:
  1. Bubbles reparsing up to parents when reparsing fails.
  2. When reparsing fails, instead of throwing, fail gracefully and highlighting the bad block in red.
  3. When text inputs are reparsed, attempt reparsing one more time without parenthesis additions, for situations like `function(a, b)`
  4. Fix text input position recovery when tabbing/shift-tabbing to work with new bubbling behavior. Previously, Droplet had located the socket relative to a nearby parent, which might now no longer exist after reparse if the reparse bubbles. This change always measures sockets relative to the entire document.


dabbler0:
This PR also makes Droplet update less frequently from its ace counterpart (it now updates only when toggling happens). It formerly updated every time ace fired a change event, but this interfered with testing, because ace fires change events at awkward times can trigger an unexpected reparse in the middle of the test.

cpirich:
Looks good.

Any possible side effects with delaying the ace updates until the next toggle?

dabbler0:
I don't think we should see anything visible from delaying updates.

src/controller.coffee
```diff
@@ -1955,8 +1968,32 @@ define ['droplet-helper',
...
+      # OR
+      #
+      # function (a) {}
+      #   -> edits [a] to [a, b]
+      #   -> reparse to a, b
+      #   -> inserts with parens to function((a, b)) {}
+      #   -> FAILS.
+      #   -> Fall back to raw reparsing the parent with unparenthesized text
+      #   -> Reparses function(a, b) {} with two paremeters.
+      #   -> Finsihed.
```
Bjvanminnen: good comments :)

src/controller.coffee
```diff
@@ -3672,6 +3690,16 @@ define ['droplet-helper',
 
     @redrawHighlights()
 
+  Editor::markBlock = (block, style) ->
+    key = @nextMarkedBlockId++
+
+    @markedBlocks[key] = {
+      model: block
+      style: style
+    }
```
Bjvanminnen: What's the reasoning for making markedBlocks an object with incrementing keys instead of just using an array?

i.e. we could do
```
@markedBlocks.push({model: block, style: style});
return @markedBlocks.length - 1

Bjvanminnen:
Looks good. Nice comments and testing :)

src/controller.coffee
dabbler0: The key is for later removal; it's an object because we want to be able to do things like `delete @markedBlocks[1]` and still have `2` point to the correct mark at `@markedBlocks[2]`. We could do this with an array, too, but an object seemed cleaner.

Bjvanminnen: makes sense

PR #67, opened 2015-06-11
Combine if/elseif chains

dabbler0:
When the parser encounters an if statement, it now checks to see if the else body is a single if statement, and if it is, does not create a new block for that if statement.


Bjvanminnen:
looks good

PR #68, opened 2015-06-11
Don't make blocks for unary operators on literal values

dabbler0:
For editing convenience, no longer make a block for values like `-1` or `+foo`.


Bjvanminnen:
looks good

PR #70, opened 2015-06-12
Change build system to use browserify instead of requirejs

dabbler0:
I've been wanting to do this change for a while. The diff looks really huge because the indent level of every source file changed by one notch, but there are no actual architecture or functionality changes in `src/`. There are meaningful changes in `Gruntfile.coffee` and in the `test` directory.

There shouldn't be any differences in the setup or build instructions. Requiring the built module with requirejs should theoretically work the same, but I haven't tested this yet; we may need to use it as a browser global or an an npm module now in other repositories.


dabbler0:
(we should probably merge this PR soon or decide not to, because it is very difficult to merge with)

davidbau:
First question - will this change anything for droplet embedders?  I.e., will there still be a droplet-full.js that is used the same way, and will ace need to be separately loaded in the same way as today?

dabbler0:
Ace will still need to be separately loaded. The method we had before of loading the script in a `script` tag and then doing `require ["droplet"], (droplet)->` will no longer work; instead we will have to do one of:

(requirejs)
```coffeescript
require ['path/to/droplet-full.js'], (droplet) ->
  new droplet.Editor() # requirejs module
```

(browser global)
```html
<script src="path/to/droplet-full.js"></script>
<script>
new droplet.Editor() //browser global
</script>
```

(npm)
```coffeescript
droplet = require './path/to/droplet-full.js'
new droplet.Editor() # npm modle
```

Bjvanminnen:
@cpirich This would result in a change in how we need to embed droplet, though it ultimately seems cleaner (just a requirejs call instead of first inject a script tag and later doing a requirejs). Can you think of any dangers in this? I guess one might be that we might hit more race conditions, since we'll actually do the load when we hit the requirejs call, whereas previously we would have loaded during the script tag injection, and retrieved from the cache during requirejs.

Bjvanminnen:
FYI for Chris, looking at this again, we can probably get it working on our side using the third option, which would likely be even cleaner for our purposes.

cpirich:
The 3rd option would presumably pull in the droplet code all the time, which we've been trying to avoid.

The 2nd method is probably the easiest. Because we can choose whether or not we want to load it using the same mechanism we do today. And then just change one line of code where we actually instantiate it.

Bjvanminnen:
Good point about not always wanting it in our package.

PR #71, opened 2015-06-15
Allow users to abort drags by replacing the block where they found it

dabbler0:
Change to `controller.coffee` to abort a drag if the user is hovering over the grayed-out block that represents the place they picked up the block. I don't know if this is a good UI decision -- thoughts?

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
     "browserify": "^10.2.4",
     "browserify-shim": "^3.8.8",
     "coffeeify": "^1.1.0",
-    "grunt": "latest",
+    "grunt": "0.4.5",
     "grunt-banner": "latest",
     "grunt-bowercopy": "latest",
     "grunt-browserify": "^3.8.0",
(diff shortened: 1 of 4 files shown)


davidbau:
Nice comments.  A good case.  Consider adding a test of drag-editing including drag-to-self to uitest.html?

davidbau:
BTW, a different usability feature that would be very nice (I see students running into it all the time):

After replacing a text socket by dragging a block on top of it, we should have the socket secretly remember the old text that was replaced.  If you pull the block back out of the socket, we should default the text to the remembered old value rather than the empty string.

Reason: it's way too easy to drop the wrong block into the wrong place.  And when you drop it into a text socket, you lose the carefully chosen text that used to be in the socket.   This would make it easy to "undo" the errant drop by dragging the block back out.

sfilman:
@davidbau I've also run into that issue a bunch and agree it would be great to fix. We've added a row to the tracking spreadsheet.

davidbau:
Nice test.  Simulating clicks isn't so bad, right?  A valuable thing to do!  LGTM.

davidbau:
@sfilman - cool, thanks.

PR #72, opened 2015-06-15
Allow command expression drops into third socket of for loop

dabbler0:
Special-case the third socket of a for loop to accept certain kinds of statement-like blocks like `a += 1`. Note that true "statements" (like `break` or `var a =` statements) still cannot and should not be able to be dropped into the third socket.


Bjvanminnen:
looks good

PR #75, opened 2015-06-16
Dropdown visuals: scroll auto and min-width to socket

dabbler0:
Two changes to the way the dropdown is shown:
  - Added 17 pixels of padding to the right and changed to `scroll:auto`. We had previously not done this because the scrollbar could sometimes cover up some text.
  - Set `min-width` to the size of the text input the dropdown is attached to, and update this when the width changes.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -24,6 +24,7 @@ CURSOR_WIDTH_DECREASE = 3
 CURSOR_HEIGHT_DECREASE = 2
 CURSOR_UNFOCUSED_OPACITY = 0.5
 DEBUG_FLAG = false
+DROPDOWN_SCROLLBAR_PADDING = 17
 
 ANY_DROP = helper.ANY_DROP
 BLOCK_ONLY = helper.BLOCK_ONLY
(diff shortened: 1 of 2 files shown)


src/controller.coffee
```diff
@@ -2204,10 +2223,8 @@ Editor::showDropdown = ->
       div.innerHTML = el.display
       div.className = 'droplet-dropdown-item'
 
-      # Match fonts
-      div.style.fontFamily = @fontFamily
-      div.style.fontSize = @fontSize
       div.style.paddingLeft = helper.DROPDOWN_ARROW_WIDTH
+      div.style.paddingRight = DROPDOWN_SCROLLBAR_PADDING
```
davidbau: Would it make sense to set this to nonzero only if the total height of the elements exceeds the max-height of the scroller?

dabbler0: I could do that, but it would require letting the browser render first (to see what the actual height is -- we can't compute this, because embedders can put arbitrary html in the dropdown choices), which means users would see a flash.

dabbler0:
Responding to earlier comment -- now checks to see if there is overflow before adding padding.

davidbau:
Nice.  LGTM.

PR #76, opened 2015-06-17
HTML mode

sakagg:
Adds support for a new mode i.e. HTML
Uses a slightly modified version of parse5 as the underlying parser.


src/languages/html.coffee
```diff
@@ -0,0 +1,681 @@
+helper = require '../helper.coffee'
+parser = require '../parser.coffee'
+
+parse5 = require '../../vendor/parse5/index.js'
```
davidbau: Now that the build is browserified, can you just add parse5 to package.json and then use require 'parse5'?

vendor/parse5/index.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
```
davidbau: Ideally, none of the parse5 files below need to be checked in now that we're using browserify.

src/languages/html.coffee
dabbler0: @sakagg -- you said that the parser was slightly modified; what are the modifications?

sakagg: https://github.com/inikulin/parse5/compare/master...sakagg:master

I've forked the repo and I can add the link to this fork in packages.json.

vendor/parse5/index.js
sakagg: Agreed
But I had to do this because of the changes I've made.

src/languages/html.coffee
davidbau: That sounds good to me.  While we wait for https://github.com/inikulin/parse5/pull/53, you can just reference your own branch in package.json and remove the files here.

sakagg:
parse5 is now loaded using npm.
Any updates?

src/languages/html.coffee
```diff
@@ -0,0 +1,684 @@
...
+      socketLevel: @getSocketLevel node
+
+  htmlSocket: (node, depth, precedence, bounds, classes) ->
+    #console.log "Adding Socket: ", JSON.stringify(bounds ? @getBounds node)
+    @addSocket
+      bounds: bounds ? @getBounds node
+      depth: depth
+      precedence: precedence
+      classes: classes ? @getClasses node
+      acccepts: @getAcceptsRule node
```
dabbler0: "acccepts" is probably a typo, although the "accepts" argument to "addSocket" is actually no longer required (we should probably update the README).

src/languages/html.coffee
```diff
@@ -0,0 +1,684 @@
...
+    line = 0
+    column = 0
+    for val, i in @text
+      @positions[i] = {'line': line, 'column': column}
+      column++
+      if val is '\n'
+        line++
+        column = 0
+    @positions[@text.length] = {'line': line, 'column': column}
+    window.positions = @positions
```
dabbler0: It seems like this was probably used for debugging and you just forgot to remove it?

src/languages/html.coffee
```diff
@@ -0,0 +1,684 @@
...
+        node.__location.end -= 3
+        @htmlSocket node, depth + 1, null
+        #node.value = node.data
+        #node.__location.start += 4
+        #@handleText node, depth + 1
+
+  isComment: (text) ->
+    text.match(/<!--.*-->/)
+
+HTMLParser.parens = (leading, trainling, node, context) ->
```
dabbler0: "trainling" -> "trailing"

src/languages/html.coffee
```diff
@@ -0,0 +1,684 @@
...
+    node.attributes.shift()
+    #console.log node.attributes
+    ###
+
+  cleanTree: (node) ->
+    if not node
+      return
+
+    if node.childNodes?
+      i = 0
```
dabbler0: Loops like this are sometimes easier to do with `filter`:
```
node.childNodes = node.childNodes.filter (child) -> child.__location?
@cleanTree(child) for child in node.childNodes
```

src/languages/html.coffee
```diff
@@ -0,0 +1,684 @@
...
+      while i < node.childNodes.length
+        @cleanTree node.childNodes[i]
+        if not node.childNodes[i].__location
+          node.childNodes = node.childNodes[0...i].concat(node.childNodes[i].childNodes || []).concat node.childNodes[i+1...]
+        else
+          i++
```
dabbler0: Loops like this are sometimes easier to do with `filter`:
```
node.childNodes = node.childNodes.filter (child) -> child.__location?
@cleanTree(child) for child in node.childNodes
```

src/languages/html.coffee
```diff
@@ -0,0 +1,684 @@
...
+      else
+        return DEFAULT_INDENT_DEPTH
+
+  handleText: (node, depth) ->
+    text = node.value.split '\n'
+    loc = {start: node.__location.start}
+    for line in text
+      loc.end = loc.start + line.length
+      @htmlSocket node, depth, null, @genBounds loc
+      loc.start = loc.end + 1
```
dabbler0: This function doesn't seem to be used anymore?

dabbler0:
Looks good aside from minor comments.

sakagg:
@dabbler0 I've fixed  the above mentioned issues and I removed all the useless comments. Most of them were for debug purposes and some were features yet to be added. I'll add them once droplet supports the features.
I also restricted dropping based on the newly added parameter `next`

src/languages/html.coffee
```diff
@@ -0,0 +1,561 @@
...
+    if blockType in allowList and blockType not in forbidList
+      return helper.ENCOURAGE
+    return helper.FORBID
+
+  switch contextType
+    when 'html'
+      if blockType is 'head'
+        if predType is 'html' and (not next or nextType is 'body')
+          return helper.ENCOURAGE
+      if blockType is 'body'
```
davidbau: frameset also works like body (only allowed right inside html)

src/languages/html.coffee
```diff
@@ -0,0 +1,561 @@
...
+        tmp++
+      line = @lines[bounds.start.line + tmp]
+      lineIndent = line[indentDepth...(line.length - line.trimLeft().length)]
+      if lineIndent.length >= DEFAULT_INDENT_DEPTH.length
+        return lineIndent
+      else
+        return DEFAULT_INDENT_DEPTH
+
+  markRoot: ->
+    @positions = []
```
davidbau: question for @sakagg and @dabbler0 - when parsing within a specific context, such as within <script>, can it parse with different rules?

For example: `<script>document.write('');</script>` - here, parsers treat everything inside as text even though there is an ``.  If editing a socket inside a `<script>`, will it reparse in script context corrrectly?

example/example-html.coffee
```diff
@@ -0,0 +1,233 @@
...
+          {"block":"<noscript></noscript>"},
+          {"block":"<template>\\n  \\n</template>"},
+          {"block":"<canvas></canvas>"},
+          {"block":"<applet code=\\"\\">\\n  \\n</applet>"},
+          {"block":"<basefont color=\\"\\" size=\\"\\" />"},
+          {"block":"<bgsound src=\\"\\" />"},
+          {"block":"<center>\\n  \\n</center>"},
+          {"block":"<command type=\\"\\" label=\\"\\" />"},
+          {"block":"<font sixe=\\"\\" color=\\"\\">\\n  \\n</font>"},
+          {"block":"<frameset cols=\\"\\">\\n  \\n</frameset>"},
```
davidbau: ... or perhaps, if you don't have rules for how drag+drop work for an element, maybe those elements shouldn't be included in the palette (such as frameset/frame).

sakagg: This problem is only for tags which are not part of HTML5 and were initially not present in the palette.
Later I added the commonly used deprecated tags to the palette but couldn't find a proper documentation of where they can be present(because they aren't part of HTML5).
So I guess I'll just go with what you mentioned.

src/languages/html.coffee
sakagg: @davidbau You might want to edit your comment since anything after `<script>` is not visible.
Moving to the problem, I was thinking of something different.
Maybe the part b/w `<script> </script>` can be handled using the JS adapter since this will be a more common use case. i.e. Handle JS code using JS adapter instead of looking for HTML tags inside js.
Once we have this, maybe we can consider checking text inside sockets of JS, and if it is valid HTML, parse is using the HTML parser?

davidbau:
OK, based on positive reviews, I'm merging this now.

PR #78, opened 2015-06-18
add newline when drag/dropping block at front of line

Bjvanminnen:
indentation

When we drag a block to the front of a line, we should add a new line + indentation to the end of the block.


dabbler0:
Looks good to me.

PR #80, opened 2015-06-19
Don't show dropdown in the palette

dabbler0:
Adds new view option and a `@paletteView` to not show dropdown in palette. Is this a good thing to do all the time or should this be configurable by the embedder?

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -224,6 +224,7 @@ exports.Editor = class Editor
 
     # Instantiate an ICE editor view
     @view = new view.View extend_ @standardViewSettings, respectEphemeral: true
+    @paletteView = new view.View extend_ @standardViewSettings, {showDropdowns: false, respectEphemeral: true}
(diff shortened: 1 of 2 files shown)


sfilman:
In general I think making things configurable is good. When we integrate
into App Lab we'll want to have it off all the time, not only to get the
space back, but also because it's a bit confusing that it's not interactive
in the toolbox.

src/controller.coffee
```diff
@@ -224,6 +224,10 @@ exports.Editor = class Editor
 
     # Instantiate an ICE editor view
     @view = new view.View extend_ @standardViewSettings, respectEphemeral: true
+    @paletteView = new view.View extend_ @standardViewSettings, {
+      showDropdowns: @options.showDropdownInPalette ? false
```
davidbau: Ah hah, got it: an option to show dropdowns at all, which is then used when drawing the palette.  Nice approach.

davidbau:
LGTM.

PR #81, opened 2015-06-19
Beginner mode for "for" loops

dabbler0:
Embedders can now specify beginner mode with a flag in the "categories" option:
```javascript
//...
modeOptions: {
  categories: {
    loops: {
      color: "green",
      beginner: true
    }
  }
}
//...
```

This will make for loops of this form mark only "b":
```javascript
for (var i = 0; i < b; i++) {
  //...
}
```

It is currently very picky about which loops it chooses to do this with; they must be exactly of the form `(/*var*/ a = x; a < b; a++)` and `a` must be the same identifier in all three expressions.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -224,6 +224,7 @@ exports.Editor = class Editor
 
     # Instantiate an ICE editor view
     @view = new view.View extend_ @standardViewSettings, respectEphemeral: true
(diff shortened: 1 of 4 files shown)


dabbler0:
Oops, hang on, this has some extraneous changes. I'll make a new PR.

PR #82, opened 2015-06-19
Beginner mode for "for" loops

dabbler0:
Embedders can now specify beginner mode with a flag in the "categories" option:
```javascript
//...
modeOptions: {
  categories: {
    loops: {
      color: "green",
      beginner: true
    }
  }
}
//...
```

This will make for loops of this form mark only "b":
```javascript
for (var i = 0; i < b; i++) {
  //...
}
```

It is currently very picky about which loops it chooses to do this with; they must be exactly of the form `(/*var*/ a = x; a < b; a++)` and `a` must be the same identifier in all three expressions.

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -413,12 +413,20 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
           @jsSocketAndMark indentDepth, node.alternate, depth + 1, 10
       when 'ForStatement'
         @jsBlock node, depth, bounds
-        if node.init?
-          @jsSocketAndMark indentDepth, node.init, depth + 1, NEVER_PAREN, null, ['for-statement-init']
-        if node.test?
-          @jsSocketAndMark indentDepth, node.test, depth + 1, 10
-        if node.update?
-          @jsSocketAndMark indentDepth, node.update, depth + 1, 10, null, ['for-statement-update']
+
+        # If we are in beginner mode, check to see if the for loop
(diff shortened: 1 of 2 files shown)


src/languages/javascript.coffee
```diff
@@ -607,6 +615,29 @@ JavaScriptParser.drop = (block, context, pred) ->
...
+  else
+    return false
+
+  return (node.test.type is 'BinaryExpression' and
+      node.test.operator is '<' and
+      node.test.left.type is 'Identifier' and
+      node.test.left.name is variableName and
+      node.update.type is 'UpdateExpression' and
+      node.update.operator is '++' and
+      node.update.argument.type is 'Identifier' and
```
davidbau: Does this accept both predecrement and postdecrment?  (If so that's good.  Predecrement is arguably the better canonical form in a loop. http://stackoverflow.com/questions/561588/what-is-more-efficient-i-or-i/561609 )

src/languages/javascript.coffee
```diff
@@ -607,6 +615,29 @@ JavaScriptParser.drop = (block, context, pred) ->
 
   return helper.DISCOURAGE
 
+isStandardForLoop = (node) ->
+  unless node.init? and node.test? and node.update?
```
davidbau: Add a comment here giving an example of what's being looked for, similar to the readme.

src/languages/javascript.coffee
dabbler0: Yep!

PR #84, opened 2015-06-23
Extra parameter 'next' in drop callback

sakagg:
Implements an extra parameter `next` in drop callback which gives information about the immediately next block, if it exists.
Can be used to ensure droppability of a block only before certain types of blocks.


dabbler0:
Looks good to me.

PR #85, opened 2015-06-23
Model cleanup

dabbler0:
Includes the following changes:
  - Token `stringify()` no longer takes `state`; will look at parent info for indent instead.
  - Added a more text-based `DropletLocation`
  - Modularized the token-count-based `serializedLocation` into the new `Location` class
  - `stringify()` no longer takes `config`; instead, the parser assigns `emptyString` to Sockets and Indents, used at stringification time.
  - `stringify()` now always gives full indents; each newline looks up the parent tree to determine indent depth.
  - Added `stringifyInPlace` with old indent behavior for use in copy-paste, etc. 
  - `moveCursorUp()` is eliminated by adding `setCursor` which can take `before` or `after`
  - Remove `append`, add non-parenting `helper.connect`
  - Added class `List` to do splicing and parenting independent of `Container`
  - Refactored some of the child-delegation code from `ContainerView` to `GenericView`
  - Added class `ListView` between `GenericView` and `ContainerView` to be able to render `List` objects. (might not be good design decision to have `ContainerView` descend... keep but consider revamping later).
  - Replaced Model `spliceIn` and `spliceOut` with `insert` and `delete` which belong to the document rather than the inserted block, and return undo operations
  - All mutations are piped through `Editor.spliceIn`, `Editor.spliceOut`, and `Editor.replace`, which in turn pipe through model `insert`, `delete` and `replace`.
  - Unified `textFocus` and `cursor` into one thing; if the `cursor` points to a `socketStart`, then we have text focus.
  - The cursor is now always a `DropletLocation`; `insert`, `delete`, and `replace` in the Model track its movements by temporarily getting a pointer to a model token when doing a mutation.
  - Unified reparsing sockets and reparsing blocks into the same concept; we can now reparse an arbitrary `List`.
  - `floatingBlocks` are now strictly listed in order of creation


davidbau:
Let's squash this history of changes - either rebase them down to one, or at least squash them down to reduce the changes that aren't important, or that don't pass tests.

davidbau:
Can you add an "architecture.md" or "design.md" that describes how the model works to people?  It's a good idea to write one now, when you're fresh off thinking through everything.

test/src/tests.coffee
```diff
@@ -113,10 +113,10 @@ asyncTest 'Get block on line', ->
...
-  strictEqual document.getBlockOnLine(3).stringify(coffee), 'console.log k', 'line 3'
-  strictEqual document.getBlockOnLine(5).stringify(coffee), 'console.log q', 'line 5'
-  strictEqual document.getBlockOnLine(7).stringify(coffee), 'console.log j', 'line 7'
+  strictEqual document.getBlockOnLine(1).stringify(), 'console.log i', 'line 1'
+  strictEqual document.getBlockOnLine(3).stringify(), 'console.log k', 'line 3'
+  strictEqual document.getBlockOnLine(5).stringify(), 'console.log q', 'line 5'
+  strictEqual document.getBlockOnLine(7).stringify(), 'console.log j', 'line 7'
   start()
 
 asyncTest 'Location serialization unity', ->
```
davidbau: consider moving this test to modelTests.coffee

test/src/tests.coffee
```diff
@@ -68,8 +68,7 @@ asyncTest 'Containers and parents', ->
   e = cont2.end
   f = cont1.end
 
-  a.append(b).append(c).append(d)
-   .append(e).append(f)
+  helper.string [a, b, c, d, e, f]
```
davidbau: consider moving this one to modelTests.coffee, along with any other tests that exercise the model without touching the container or view.

davidbau:
modelTests.coffee is a bit brief, given that you need the model to be correct for everything else in the system to function.  Consider writing more tests, for example, for each explanation in architecture.md, write a brief test to exercise it.

dabbler0:
Updates: added a section on model architecture to `DESIGN.md`, moved Model tests from `tests.coffee` to `modelTests.coffee` and added a simpler undo/redo unit test.

PR #86, opened 2015-06-25
Separate paletteWrapper and paletteElement.

davidbau:
Previously these two controller members referred to the same
element.  Now they are different, one nested in the other.

Now paletteWrapper deals with the sizing of the overall palette pane
but not the clipping or the containing, and paletteElement contains
all the children and does the clipping.  This allows embedders
such as Pencil Code to decorate the palette pane by attaching elements
to paletteWrapper (such as mode- or palette-switching buttons and menus)
without interfering with the working of the palette.

This is a no-op for most users of pencil code.

### DIFF
--- a/css/droplet.css
+++ b/css/droplet.css
@@ -69,13 +69,18 @@
   position: absolute;
   top: 0; bottom: 0; width: 300px;
   max-width: 40%;
-  overflow: hidden;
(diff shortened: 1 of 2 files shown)


dabbler0:
Looks good to me.

PR #87, opened 2015-06-26
Gutter resize fix

dabbler0:
Resize/reposition the main canvas when the gutter resizes to avoid this overlapping issue:


Bjvanminnen:
looks good

PR #88, opened 2015-06-26
Add classes and droppability for comment blocks

dabbler0:
Allow comment blocks to be dropped in indent contexts, and don't add parentheses or semicolons to them.


Bjvanminnen:
looks good

PR #91, opened 2015-06-30
Remove 'accepts' from README

sakagg:
Removes `accepts` from `addSocket` in README.md since the feature has been removed from droplet.


davidbau:
Thanks!

PR #94, opened 2015-06-30
Floating block assembly

dabbler0:
Allow users to assemble blocks floating in space. Includes the following architecture changes:
 - Undo stack operations now come with an associated number indicating the index of the document to which they apply (0 is the main tree, subsequent ones are floating blocks)
 - The cursor comes with a similar associated index

The lasso select will only select in one document at a time; when nothing is selected, the user can move the lasso over another document to select in it.

This change also includes some extra testing and bug fixes for cursor operations that came up while I was implementing this.


src/controller.coffee
```diff
@@ -424,25 +429,108 @@ Editor::redrawMain = (opts = {}) ->
...
+    endWidth = @mainCtx.measureText(@mode.endComment).width
+    for record in @floatingBlocks
+      blockView = @view.getViewNodeFor record.block
+      blockView.layout record.position.x, record.position.y
+
+      rectangle = new @view.draw.Rectangle(); rectangle.copy(blockView.totalBounds)
+      rectangle.x -= GRAY_BLOCK_MARGIN; rectangle.y -= GRAY_BLOCK_MARGIN
+      rectangle.width += 2 * GRAY_BLOCK_MARGIN; rectangle.height += 2 * GRAY_BLOCK_MARGIN
+
+      bottomTextPosition = blockView.totalBounds.bottom() - blockView.distanceToBase[blockView.lineLength - 1].below - @fontSize
```
davidbau: this is a lot of complexity to put inline in the main function.  Can you factor out drawing each floating block into its own function, perhaps?

src/draw.coffee
```diff
@@ -202,20 +207,24 @@ exports.Draw = class Draw
...
 
       _clearCache: ->
+        @_cacheFlag = true
         if @_cacheFlag
-          minX = minY = Infinity
-          maxX = maxY = 0
-          for point in @_points
-            minX = min minX, point.x
-            maxX = max maxX, point.x
+          if @_points.length is 0
```
davidbau: Didn't read this too carefully yet - but a general question - why does any caching logic have to change due to the floating assembly change?

dabbler0: This was a general bug fix where the bounding box of an empty path would be NaN/NaN instead of 0/0 I discovered when I was making the surrounding dotted-line paths for the floating blocks.

dabbler0:
Update -- Factored out drawing a floating block, and added some comments in `draw.coffee`

PR #95, opened 2015-06-30
Simple recovery from coffee quoting errors in a socket.

davidbau:
This change allows coffeescript to recover from unescaped quotes entered in a socket.

If you enter `'What's up'` in a coffeescript socket, that string will misparse, but it will be caught as having quote characters on the edges, and it will be transformed into a correctly-escaped string by the parsing recovery logic.


davidbau:
Anthony reviewed separately.  Merging.

PR #97, opened 2015-07-01
Add line bounding box API

dabbler0:
`Editor.getLineBoundingBox(row)` where `row` is 0-indexed will return an instance of Droplet's `draw.Rectangle`, which has properties `x`, `y` ,`width`, `height` and methods `right()` and `bottom()`.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -4016,8 +4016,10 @@ Editor::viewportDimensions = ->
 # =================
 Editor::getLineMetrics = (row) ->
   viewNode = @view.getViewNodeFor @tree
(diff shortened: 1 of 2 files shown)


src/controller.coffee
```diff
@@ -4012,6 +4012,11 @@ Editor::viewportDimensions = ->
     height: @mainCanvas.height
   }
 
+# LINE LOCATION API
+# =================
+Editor::getLineBoundingBox = (row) ->
```
davidbau: maybe rename to getLineDrawingMetrics

and add things like distance-to-text baseline.

PR #100, opened 2015-07-09
HTML Bugfixes

sakagg:
Bugfixes.
1) Earlier if an attribute value was empty, it wasn't editable in block mode. Now you can edit it.
2) EDIT: Fixed another bug where some blocks were not reparsing corrrectly due to lack of context information.
3) Little cleanup on parser.coffee#L221


dabbler0:
Looks good to me.

PR #101, opened 2015-07-10
Dotted lines

dabbler0:
Replace the gray block outline with dotted lines.


dabbler0:
This PR fails because of a test bug that was fixed in `remember-socket-floating`. Submitting small changes directly to `floating-assembly` and merging `remember-socket-floating`.

PR #107, opened 2015-07-15
SVG [WIP]

dabbler0:
Render with SVG instead of canvas, for resolution and eventually for better mouse events.

TODO:
  - text input for some reason doesn't let you select the text input thing sometimes, instead reparses immediately
  - melt/freeze animations should be aware of empty socket values to avoid pop
  - make the cursor always correct during lasso selects
  - top-down caching for cache removal logic -- if version number is the same, look no further

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -69,7 +69,7 @@ module.exports = (grunt) ->
     browserify:
       build:
         files:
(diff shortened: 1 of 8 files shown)


dabbler0:
Closed in favor of `svg-grouping` #172

PR #108, opened 2015-07-17
HTML colour fixes

sakagg:
Assign colours to nodes based on the category and the node colour support is much wider.
The colour of the nodes with make it easy for the children to remember what node belongs to which category.

### DIFF
--- a/src/languages/html.coffee
+++ b/src/languages/html.coffee
@@ -5,142 +5,151 @@ parse5 = require 'parse5'
 
 ATTRIBUTE_CLASSES = ['#attribute']
 
-COLORS = {
+TAGS = {
   #Metadata
-  '#documentType': 'lightblue'
-  html: 'lightblue'
-  head: 'lightblue'
-  body: 'lightblue'
-  title: 'lightblue'
-  link: 'lightblue'
-  meta: 'lightblue'
-  style: 'lightblue'
-  script: 'lightblue'
-  base: 'lightblue'
+  '#documentType': {category: 'metadata'}
+  html: {category: 'metadata'}
+  head: {category: 'metadata'}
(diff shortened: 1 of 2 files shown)


src/languages/html.coffee
```diff
@@ -6,41 +6,141 @@ parse5 = require 'parse5'
...
-  'p': 'deeppurple',
-  'script': 'orange',
-  'select': 'indigo',
-  'strong': 'yellow',
-  'table': 'lightgreen',
-  'td': 'lightblue',
-  'title': 'green',
-  'tr': 'bluegrey',
-  'ul': 'blue'
+  #Metadata
```
davidbau: Let's go ahead and make this more similar to the configuration in coffee.coffee.

  * Make it do an indirection through a "categories" option that you can override in modeOptions.  So that you can change the color of whole categories by just changing modeOptions.categories.

  * For each tag, instead of just specifying a color, specify a whole object that in the future can include a dropdown option, like  `{ category: 'content', dropdown: [['href=""', 'name=""']] }`.

  * Don't worry about adding the dropdown option here yet, but that would be the next thing to do in a future PR - based on attribute statistics, add the most common attributes as choices.

sakagg: On it!
Can you tell me why is there a list of lists for dropdown?

davidbau: Dropdown is organized as a menu for each argument position (for example, if the first argument of a function is a color and the second argument is a number, then you would have a different set of choices for each position).

In HTML, we would want to have a different organizing scheme, since every attribute has the same choices. We'll have to look into dropdown support later and see what's needed for that.

davidbau:
Looks great.

PR #113, opened 2015-07-29
Don't insert text blocks when Return is pressed if the Control or Command keys are held down

gonfunko:
This fixes an issue in PencilCode where the shortcut to run the current program (Command-Return, or Control-Return on Linux/Windows) would also inadvertently add a new blank text block. Shift-Return was already excluded from inserting a new block.


davidbau:
LGTM

PR #116, opened 2015-07-30
Fix string autoescape and add test

dabbler0:
Fixes a feature that has been broken for a while or may never have worked, where in CoffeeScript mode if students write a quote inside a string it will automatically escape it. Adds a test for this feature as well.

### DIFF
--- a/src/languages/coffee.coffee
+++ b/src/languages/coffee.coffee
@@ -180,13 +180,13 @@ exports.CoffeeScriptParser = class CoffeeScriptParser extends parser.Parser
 
   markRoot: ->
     # Preprocess comments
-    do @stripComments
-
     retries = Math.max(1, Math.min(5, Math.ceil(@lines.length / 2)))
     firstError = null
     # Get the CoffeeScript AST from the text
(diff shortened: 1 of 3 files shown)


src/languages/coffee.coffee
```diff
@@ -1032,11 +1032,11 @@ fixCoffeeScriptError = (lines, e) ->
 fixQuotedString = (lines) ->
   line = lines[0]
   quotechar = if /^"|"$/.test(line) then '"' else "'"
-  if line.getCharAt(0) is quotechar
+  if line.charAt(0) is quotechar
     line = line.substr(1)
-  if line.getCharAt(line.length - 1) is quotechar
-    line = line.substr(0, line.length -1)
-  return quoteAndCEscape fixQuotedLine line
```
davidbau: Hm, seems like it was previously dead code!

This should read: quoteAndCEscape looseCUnescape(line), quotechar

davidbau: OK, I've pushed a fix.

PR #118, opened 2015-07-31
Add a socket when there are no params in a function call

dabbler0:
This affects JavaScript and CoffeeScript mode. This will not be done in JavaScript if there are intervening comments or line breaks between the function variable and parentheses.

### DIFF
--- a/src/languages/coffee.coffee
+++ b/src/languages/coffee.coffee
@@ -553,6 +553,28 @@ exports.CoffeeScriptParser = class CoffeeScriptParser extends parser.Parser
             # In the 'beginner' case of a simple method call with a
             # simple base object variable, let the variable be socketed.
             @csSocketAndMark node.variable.base, depth + 1, 0, indentDepth
+
+          if node.args.length is 0 and not node.do
+            variableBounds = @getBounds(node.variable)
+            start = {
+              line: variableBounds.end.line
+              column: variableBounds.end.column + 1
+            }
+            end = {
+              line: start.line
+              column: start.column
+            }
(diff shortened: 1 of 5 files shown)


src/languages/coffee.coffee
```diff
@@ -553,6 +553,28 @@ exports.CoffeeScriptParser = class CoffeeScriptParser extends parser.Parser
             # In the 'beginner' case of a simple method call with a
             # simple base object variable, let the variable be socketed.
             @csSocketAndMark node.variable.base, depth + 1, 0, indentDepth
+
+          if node.args.length is 0 and not node.do
+            variableBounds = @getBounds(node.variable)
```
davidbau: Add a comment that you are assuming an open parentheses immediately after the node.variable.

src/languages/coffee.coffee
```diff
@@ -566,6 +588,8 @@ exports.CoffeeScriptParser = class CoffeeScriptParser extends parser.Parser
               # Inline function definitions that appear as the last arg
               # of a function call will be melded into the parent block.
               @addCode arg, depth + 1, indentDepth
+            else if index is 0
```
davidbau: check that arg list length <= 1 as well.

davidbau:
Hm, not ready to merge this change yet - even when the flag is on, the sockets for zero-argument functions should only be shown on user-defined functions, not built-in-recognized ones.

davidbau:
@dabbler0 - I'd like to merge this - it fixes a key issue that's blocking one of our teachers.  Look OK?

PR #123, opened 2015-08-10
CSS Mode

sakagg:
Adds an additional CSS mode to droplet.


davidbau:
Hi Saksham - are you pretty happy with CSS mode?  If so I'll get ready to merge it.

sakagg:
One thing before you merge it. Do we want to keep all properties of the same color, or color code them as well?
Because that is half way implemented.

davidbau:
My inclination is to color-code the properties, since I think CSS beginners will spend most of their time exploring those (but only using very simple selectors)

sakagg:
I think this is ready now. Let me know what you think.

PR #129, opened 2015-10-15
Request to merge all changes from code-dot-org to master

cpirich:
One some simpler tutorials, we don't want to confuse students by allowing them to drop blocks in the middle of lines of text. Now, if you set the `dropIntoAceAtLineStart` property in the `options` block to true, the dropping code will always insert the blocks as a new line of code - at the proper indentation level.

Also, created a new `palettetoggledone` event that fires after the animation that follows `enablePalette` is complete.


cpirich:
Re: commit https://github.com/droplet-editor/droplet/commit/906cc325d48833891f9e683dddc3a41c5011666a
The change to make dropdowns functional within the palette is a little complex. The palette blocks don't belong to a "droplet document", so quite a bit of code had to be touched. Either because code assumed there would always be a document or because code wanted to create Undo operations, which don't make sense in this case.

dabbler0:
Dropdowns in the palette -- in the future, if we end up wanting to add more mutability to the palette, we might just want to wrap each palette block in its own Document (like the floating blocks) instead of special-casing them. This should be fine for dropdowns, though; we can clean it up later if we end up wanting more mutability.

I think `@start` is always a `DocumentStartToken` only because `getFromTextLocation` happens to only called on `Document`s. It could theoretically be called on other kinds of containers (blocks, sockets, etc.). Maybe we should move `getFromTextLocation` into the `Document` class instead, so that it becomes unambiguous.

Otherwise, all of these changes look good to me.

dabbler0:
I think it would be best to merge this branch now; there are some structure changes (dropdowns and `getFromTextLocation`) we might want to do, but these aren't unique to this branch, and this branch contains enough bug fixes that we've been branching off of it instead of master. We can restructure sometime down the road if it becomes necessary.

Merging now.

PR #130, opened 2015-10-22
Prevent reparsing of function parameters socket in JavaScript mode

joshlory:
From @dabbler0: 
> I think the best way to fix this bug would be to set a flag on that parameter socket saying always to bubble up reparses to its parent (never reparse it by itself). In javascript.coffee, set the flag on that socket through the "parseContext" property of the socket (make it something like "NO_REPARSE"). Then, in Editor::reparse in controller.coffee, look for that value of "parseContext", and always bubble up when you see it.

Sockets didn't have a `parseContext` so I added it to the model and to `@addSocket`.  When `parseContext == 'NO_REPARSE'` the reparse of the socket contents is skipped, and the parent is reparsed as a whole.

`'NO_REPARSE'` context is set for the JavaScript FunctionDeclaration parameters socket.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -2153,7 +2153,8 @@ Editor::reparse = (list, recovery, updates = [], originalTrigger = list) ->
     return if list.start.next is list.end
 
     originalText = list.textContent()
-    @reparse new model.List(list.start.next, list.end.prev), recovery, updates, originalTrigger
+    unless list.parseContext = 'NO_REPARSE'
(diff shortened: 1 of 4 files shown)


joshlory:
I need to fix the unit test.

The other part I don't fully understand is why reparsing the parent succeeded before when the socket is wrapped in parentheses.  The following should fail to reparse (triggering a fallback to `originalText`), but doesn't:
```javascript
function ((a)) {}
```

joshlory:
See https://github.com/droplet-editor/droplet/pull/131 for a better root-cause fix.  Closing this PR.

PR #131, opened 2015-10-23
Restore update locations when a socket reparse fails

joshlory:
Prior to this change, socket reparse would modify the `updates` array of locations, but not restore the original values if the parent fails to parse with the new socket value. This would cause `Container#getFromLocation` to throw after falling off the end of the linked list.

The fix is to save the update locations before attempting to parse the socket as a new block, and restore them if the parent reparse fails.

This is a better fix for the JavaScript function sockets bug than https://github.com/droplet-editor/droplet/pull/130, which only treats the symptom but not the root cause.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -2153,12 +2153,17 @@ Editor::reparse = (list, recovery, updates = [], originalTrigger = list) ->
     return if list.start.next is list.end
 
     originalText = list.textContent()
+    originalUpdates = updates.map (location) ->
+      count: location.count, type: location.type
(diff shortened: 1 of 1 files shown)


Bjvanminnen:
Do you have enough understanding now to update the comment at the beginning of the method?
```
   # Don't reparse sockets. When we reparse sockets,
   # reparse them first, then try reparsing their parent and
   # make sure everything checks out.

Bjvanminnen:
Changes lgtm. Good job tracking this down.

PR #132, opened 2015-10-24
Fix cmd+x having `pressedVKey = true`

joshlory:
Cut and paste could get the editor into a bad state because `pressedVKey` wasn't always being cleared. Pressing `cmd+v` then `cmd+x` would hit the paste code path and throw.


Bjvanminnen:
lgtm

PR #133, opened 2015-10-30
Enable custom builds with only one language

joshlory:
With `grunt build uglify` droplet-full.min.js is ~1.57 mb.
With `LANGUAGE=javascript grunt build uglify` droplet-full.min.js is ~0.31 mb (1/5 the size).

This is important during Hour of Code where every byte across the wire counts!

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -8,7 +8,7 @@ path = require 'path'
 fs = require 'fs'
 
 if process.env.LANGUAGE
-  ignoredLanguages = ['./src/languages/!(' + process.env.LANGUAGE + ').coffee']
+  ignoredLanguages = ["./src/languages/!(#{process.env.LANGUAGE}).coffee"]
 
 serveNoDottedFiles = (connect, options, middlewares) ->
   # Avoid leaking .git/.svn or other dotted files from test servers.
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -7,6 +7,9 @@ livereload = require 'tiny-lr'
 path = require 'path'
 fs = require 'fs'
 
+if process.env.LANGUAGE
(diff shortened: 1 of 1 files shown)


Gruntfile.coffee
```diff
@@ -75,6 +78,7 @@ module.exports = (grunt) ->
         files:
           'dist/droplet-full.js': ['./src/main.coffee']
         options:
+          ignore: ignoredLanguages
```
Bjvanminnen: Presumably when LANGUAGE is not set, ignoredLanguages is undefined and coffeescript just handles this all fine?

Gruntfile.coffee
```diff
@@ -7,6 +7,9 @@ livereload = require 'tiny-lr'
 path = require 'path'
 fs = require 'fs'
 
+if process.env.LANGUAGE
+  ignoredLanguages = ['./src/languages/!(' + process.env.LANGUAGE + ').coffee']
```
Bjvanminnen: It looks like coffeescript supports the following syntax, which is probably a little cleaner.
```
ignoredLanguages = ["./src/languages/!#{process.env.LANGUAGE}.coffee"]
```

Gruntfile.coffee
joshlory: Yeah I *think* I'm doing this the right way.  Coffeescript doesn't have `var` statements, and `if` statements don't create a new scope.  The generated JavaScript will look something like:
```javascript
var ignoredLanguages;

if (process.env.LANGUAGE) {
  ignoredLanguages = ["./src/languages/!(" + process.env.LANGUAGE + ").coffee"];
}
```

Gruntfile.coffee
joshlory: Good call, fixed in 498369d7791faf2e071f1d0e4f20515a7e28c314!

Bjvanminnen:
lgtm

PR #134, opened 2015-10-31
Unify `@opts.zeroParamFunctions` behavior

joshlory:
Commit https://github.com/droplet-editor/droplet/commit/62b40fb727cd5a938fb49342b19c7136660decf7 added a click-target for user-defined functions with no parameters, so new params could be added in block mode.  Commit https://github.com/droplet-editor/droplet/commit/bb283afba434d38aee0201f420b72cfc939fa6d6 made that option configurable, but seems to have changed the default behavior to `false`.

This change re-enables adding params to user-definied functions with no params. It also updates function calls to also respect the `@opts.lockZeroParamFunctions` flag.

The default value of the modified flag is `false`, which doesn't lock zero-parameter functions (new parameters may be added by clicking inside the parentheses).

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -371,7 +371,7 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
             classes: ['no-drop']
             empty: ''
           }
-        else if @opts.zeroParamFunctions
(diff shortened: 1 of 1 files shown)


Bjvanminnen:
lgtm

PR #135, opened 2015-11-02
Clicking the palette scrollbar shouldn't trigger a mousedown event

### LINKED ISSUE #104: Clicking on scrollbar ends up dragging elements underneath
If you drag the scrollbar of the main editing region, the clicks end up going through to elements that are underneath the scrollbar.  To repro,

  1. create a page with big blocks that go past the right and bottom edges, like this one: http://david.pencilcode.net/edit/about.html#blocks=1

  2. then click on the scrollbar where some block is underneath, and drag the scrollbar.

  3. observe: hovering highlights the block, and click-drag drags the block.

expect: the click shouldn't go through the scrollbar.

joshlory:
Fixes #104.  Bind events separately for `.droplet-palette-header` and `.droplet-palette-scroller-stuffing` so clicking the palette scrollbar doesn't trigger the callback.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -245,8 +245,8 @@ exports.Editor = class Editor
     for eventName, elements of {
         keydown: [@dropletElement, @paletteElement]
         keyup: [@dropletElement, @paletteElement]
-        mousedown: [@dropletElement, @paletteElement, @dragCover]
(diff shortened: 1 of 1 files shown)


Bjvanminnen:
lgtm

PR #136, opened 2015-11-02
Cleanup

joshlory:


.gitignore
```diff
@@ -7,3 +7,5 @@ example/*.js
 example/*.map
 test/*.js
 test/*.map
+.DS_Store
+.idea/
```
Bjvanminnen: should this instead be in a global .gitignore file instead of project specific?

PR #139, opened 2016-01-25
Fix off-by-one bug in the if/else add-button

joshlory:
Incorrect JS was generated when pressing the '+' button on an else statement with no space before the bracket.

Before: `else{` -> `els if (__) e{`
After: `else{` -> `else if {`

The incorrect JS would fail to parse, throwing an error in acorn.js.  To the end-user it would appear as if the '+' button had no effect.

This bug was hidden for else statements with a space because `elseLocation` could be moved back one character (across the space) and still not impact the 'else' token.  The bug is only triggered when the space is missing.  Both tokens are trimmed after splitting so there is no change in the generated output.

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -772,7 +772,7 @@ JavaScriptParser.handleButton = (text, button, oldBlock) ->
 
     if elseLocation?
       lines = text.split('\n')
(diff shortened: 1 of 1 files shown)


joshlory:
Travis CI build fails even after reverting this change.

Bjvanminnen:
lgtm. nice find

PR #141, opened 2016-01-26
Ensure simple for loops don't have a block in node.test.right socket

joshlory:
This bug fix is for when the 'beginner' config option is set.  Without this change the following JS throws a parser error when switching from text to block mode:
```javascript
var items = [4, 5, 6];
for (var i = 0; i < items.length; i++) {
  //
}
```
This is because the beginner for loop doesn't expect a nested block inside the test condition.

I've updated `isStandardForLoop` to also check that the right-side test is a Literal or Identifier.  If it's not, Droplet falls back to parsing the node as a non-beginner for loop.

---

Beginner for loop:
beginner-for-loop

Non-beginner for loop:
non-beginner-for-loop

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -736,6 +736,7 @@ isStandardForLoop = (node) ->
       node.test.operator is '<' and
       node.test.left.type is 'Identifier' and
       node.test.left.name is variableName and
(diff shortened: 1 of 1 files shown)


Bjvanminnen:
lgtm. will merge once travis reports green

PR #142, opened 2016-02-10
When unmarking, only redraw highlights canvas

joshlory:
This gives a ~10x speed improvement when clearing highlights.


cpirich:
Looks good, but it looks like a separate, existing bug is that `unmark` doesn't redraw at all. Do you want to fix this at the same time?

joshlory:
Good catch, fixed in d20ac9ba835569abf9671e75272b741ff027de10.

cpirich:
LGTM - should I merge?

joshlory:
Yes, thank you!

PR #144, opened 2016-02-10
Amend https://github.com/droplet-editor/droplet/pull/143

joshlory:
Feedback from @dabbler0: `@start` is always a `DocumentStartToken`. (See https://github.com/droplet-editor/droplet/pull/143)

### DIFF
--- a/src/model.coffee
+++ b/src/model.coffee
@@ -696,11 +696,8 @@ exports.Container = class Container extends List
   # ## getFromTextLocation ##
   # Given a TextLocation, find the token at that row, column, and length.
   getFromTextLocation: (location) ->
-    head = @start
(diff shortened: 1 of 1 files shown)


src/model.coffee
```diff
@@ -696,12 +696,9 @@ exports.Container = class Container extends List
   # ## getFromTextLocation ##
   # Given a TextLocation, find the token at that row, column, and length.
   getFromTextLocation: (location) ->
-    head = @start
+    head = @start.next
```
Bjvanminnen: Worth a comment about why we're going to start.next?

joshlory: Fixed in a55ce072ede28e31d1d254c463bd5a598e97d659.

PR #146, opened 2016-02-18
ability to scroll cursor to end of document

Bjvanminnen:
Method to scroll to end of document. Will be called from App Lab after we append code to the end of the document.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -2869,12 +2869,14 @@ Editor::scrollCursorIntoPosition = ->
 # Moves the cursor to the end of the document and scrolls it into position
 # (in block and text mode)
 Editor::scrollCursorToEndOfDocument = ->
-  pos = @tree.end
-  while pos && !@validCursorPosition(pos)
-    pos = pos.prev
-  @setCursor(pos)
-  @scrollCursorIntoPosition()
-  @aceEditor.scrollToLine @aceEditor.session.getLength()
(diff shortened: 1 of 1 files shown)


joshlory:
This feels like a very specific API.  I wonder if it would be better to have this method take a line number (with -1 for the last line of the document, -2 for the penultimate line, etc).

Bjvanminnen:
I'm somewhat wary of making a more generalized API without there being a clear need. Also `moveAndScrollCursorToLine(-1)` feels less clear than `scrollCursorToEndOfDocument()`, so we'd be sacrificing clarity for generalizability that it's not clear we need imo.

src/controller.coffee
```diff
@@ -2866,6 +2866,17 @@ Editor::scrollCursorIntoPosition = ->
...
 
+# Moves the cursor to the end of the document and scrolls it into position
+# (in block and text mode)
+Editor::scrollCursorToEndOfDocument = ->
+  pos = @tree.end
+  while pos && !@validCursorPosition(pos)
+    pos = pos.prev
+  @setCursor(pos)
+  @scrollCursorIntoPosition()
+  @aceEditor.scrollToLine @aceEditor.session.getLength()
```
joshlory: Do we need to scroll both text and block mode, or can we detect which mode they're in and scroll only that mode?

joshlory:
That makes sense.  LGTM

PR #150, opened 2016-03-18
Limit scope of redraws for highlights and cursors

joshlory:
There were a few places where Droplet redraws more than it needs to.  These changes are part of the focus-socket-on-mouseup changes.  Moving them to a separate PR to merge them early.


Bjvanminnen:
lgtm

PR #151, opened 2016-03-25
Firefox doesn't have a global event object

joshlory:
This fixes dragging `if` blocks into `for` loops, but there may be other instances of this bug.  Root cause is that `window.event` is undefined in Firefox.  In other browsers it's the event currently being handled.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -1369,7 +1369,7 @@ hook 'mousemove', 1, (point, event, state) ->
     # Redraw the main canvas
     @redrawMain()
 
-Editor::getClosestDroppableBlock = (mainPoint, event) ->
+Editor::getClosestDroppableBlock = (mainPoint, isDebugMode) ->
   best = null; min = Infinity
 
   if not (@dropPointQuadTree)
@@ -1381,7 +1381,7 @@ Editor::getClosestDroppableBlock = (mainPoint, event) ->
     w: MAX_DROP_DISTANCE * 2
     h: MAX_DROP_DISTANCE * 2
   }, (point) =>
-    unless (point.acceptLevel is helper.DISCOURAGE) and not event?.shiftKey
+    unless (point.acceptLevel is helper.DISCOURAGE) and not isDebugMode
       # Find a modified "distance" to the point
       # that weights horizontal distance more
       distance = mainPoint.from(point)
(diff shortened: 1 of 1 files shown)


src/controller.coffee
```diff
@@ -214,6 +214,8 @@ exports.Editor = class Editor
     # ## Tracker Events
     # We allow binding to the tracker element.
     dispatchMouseEvent = (event) =>
+      # Firefox doens't have a global event object
+      window.event = event
```
Bjvanminnen: Googling seems to imply that window.event is actually IE only (https://developer.mozilla.org/en-US/docs/Web/API/Window/event). Would the better fix be to make it so that nobody depends on event being in the global namespace?

joshlory: Yep, this is the long-term fix.  Should I do that fix now before check-in?  Adding this change for now will fix the user-facing bug in the meantime.

Bjvanminnen: How come this isn't a problem in Chrome (where window.event is not a global?)

joshlory: It does seem to be global in Chrome:
```javascript
document.addEventListener('click', function () { console.log(window.event); })
```

Bjvanminnen: ahh right, i had just typed `window.event` and it was undefined, but that's of course because I wasn't in the process of dealing with an event. Interesting the mozilla article from above says it's IE only.

Modifying the global namespace as a workaround feels a little bit scary to me. I worry that by committing this fix, we make finding the underlying issue (dependence on the global) something we're less likely to actually do, and that we could be introducing other subtle differences in behavior (tho i'm not sure that's too big a worry in this case).

I'd feel better if we tracked down the real bug, and maybe if that proves particularly difficult commit this workaround. What do you think?

joshlory:
Updated with a better fix.

Bjvanminnen:
This definitely feels better :)

However, it feels strange that getClosestDroppableBlock has to know whether or not the shift key is pressed.

If I read this right, I think when the shift key is pressed this will always return null? If so, would it make more sense for callers to check the shiftKey (and refrain from calling when pressed)?

joshlory:
I think it has to do with the debug feature in Droplet.  When the shift key is pressed, Droplet will offer drop targets even if `point.acceptLevel is helper.DISCOURAGE`.

joshlory:
Looking at the test failures now.

src/controller.coffee
```diff
@@ -1381,7 +1381,7 @@ Editor::getClosestDroppableBlock = (mainPoint) ->
     w: MAX_DROP_DISTANCE * 2
     h: MAX_DROP_DISTANCE * 2
   }, (point) =>
-    unless (point.acceptLevel is helper.DISCOURAGE) and not event.shiftKey
+    unless (point.acceptLevel is helper.DISCOURAGE) and not event?.shiftKey
```
Bjvanminnen: I really wish JS had this syntax

Bjvanminnen: It still feels a little weird to me that getClosestDroppableBlock needs an event. What if we instead make this signature `Editor::getClosestDroppableBlock = (mainPoint, isDebugMode)` and then the caller is `@getClosestDroppableBlock(mainPoint, event?.shiftKey)`?

Bjvanminnen:
lgtm

PR #155, opened 2016-04-06
Redraw when an add- or subtract-button click is handled

joshlory:
Prior to https://github.com/droplet-editor/droplet/pull/150 the redraw was triggered indirectly via `mousedown -> clearLassoSelection() -> redrawHighlights() -> redrawMain()`.

This fixes the bug where pressing the '+' button on an `if` block appears to have no effect until something else happens to trigger a `redrawMain()`.


cpirich:
LGTM

PR #156, opened 2016-04-21
expose changePaletteGroup method, allow groups to have ids

cpirich:
* We want to allow certain levels in our curriculum to start up droplet with the palette group as something other than the default group (which is currently always the first one).
* Allow the supplied `@paletteGroups` object to optionally specify an `id` for each group. When present, it will be used to generate a DOM id for the group header (e.g. if `math` is the id, the header DOM id will be `droplet-palette-group-header-math`). This is useful for assigning tooltips to the header elements.
* Exposed a new `changePaletteGroup(group)` method and changed the existing code to use that method. The `group` parameter will be used to match from the `@paletteGroups` array - it can match the group object itself, the `name`, or the `id` of the group.
* The `@paletteGroups` object (supplied at editor creation) is now extended to retain the `header` element and the `parsedBlocks` array since they are no longer available through a closure.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -1834,8 +1834,10 @@ Editor::setPalette = (paletteGroups) ->
         paletteHeaderRow.style.height = 0
 
     # Create the element itself
-    paletteGroupHeader = document.createElement 'div'
+    paletteGroupHeader = paletteGroup.header = document.createElement 'div'
     paletteGroupHeader.className = 'droplet-palette-group-header'
+    if paletteGroup.id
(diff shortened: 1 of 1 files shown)


cpirich:
@Bjvanminnen

src/controller.coffee
```diff
@@ -1894,6 +1876,37 @@ Editor::setPalette = (paletteGroups) ->
   @resizePalette()
   @resizePaletteHighlight()
 
+Editor::changePaletteGroup = (group) ->
```
Bjvanminnen: Could use a comment describing what this function does

Bjvanminnen:
lgtm

PR #157, opened 2016-05-10
JavaScript parameter buttons

cpirich:
(1) Update `code-dot-org` branch with latest commits from master
(2) Change `addButton` and `subtractButton` properties to hold a character value (rather than a `boolean`), so they can render as something other than `+` and `-`
(3) Change the order of the `addButton` and `subtractButton` when both render so that `subtractButton` is first (on the left)
(4) Allow known functions to specify that they'd like to have `minArgs`, `maxArgs`, or both. If either or both is present, a click on the `addButton` will add an extra parameter socket to the function call (unless it would exceed `maxArgs`). Similarly, a click on the `subtractButton` will remove the last parameter from the function call (unless it would reduce below `minArgs`). This is useful if certain functions have a variable number of arguments.
(5) Fixed a bug related to functions with only argument. An earlier commit (https://github.com/droplet-editor/droplet/pull/118/commits/4ff23d571255b624b7523527196c34abf3aa054a) made it possible to delete the first parameter of a function call by deleting the text from a socket and pressing Enter. However, that commit also made it impossible to create a function with a single parameter socket expressed with the `__` representing an empty parameter (e.g. `myFunction(__)`). I reverted that change for JavaScript only. You can no longer delete the only parameter from a 1 parameter function call with the keyboard and have it turn into a 0 parameter function call, but you can properly use the `__` in the palette and it also enables the `addButton` to work as expected when expanding from a 0 parameter function call to a 1 parameter function call.
(6) Moved the parser's `handleButton` method to belong to a parser instance object. This helps so that it can check for known functions with `@lookupKnownName`
(7) Add a mode option called `paramButtonsForUnknownFunctions` - when this is `true`, add and subtract buttons will appear on function calls for "unknown" functions.

_NOTE: the visuals for the `addButton` and `subtractButton` probably need some work. Ideally, these buttons would appear just before the right parentheses, but that will take quite a bit more work_

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "grunt-cli": "latest",
     "grunt-contrib-coffee": "latest",
     "grunt-contrib-concat": "latest",
-    "grunt-contrib-connect": "1.0.0",
+    "grunt-contrib-connect": "latest",
     "grunt-contrib-cssmin": "latest",
     "grunt-contrib-qunit": "<=0.5.2",
     "grunt-contrib-uglify": "latest",
@@ -56,6 +56,6 @@
     "acorn": "^1.2.2",
     "sax": "^1.1.1",
     "antlr4": "latest",
-    "parse5": "1.5.1"
+    "parse5": "latest"
   }
 }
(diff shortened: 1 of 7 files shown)


cpirich:
@Bjvanminnen @joshlory

Bjvanminnen:
travis-ci failures expected?

Bjvanminnen:
changes lgtm, tho i have minimal expertise here

cpirich:
travis-ci failures seem unrelated (all in the html area - related to changes merged from master). This will need to be sorted out when merged back to master. I can reproduce the test failures locally. But our build only includes the JavaScript parser, so it doesn't seem to be blocking.

PR #159, opened 2016-05-11
Single-line comment changes

cpirich:
* Change `constructHandwrittenBlock` in the `isComment` case to call a new `indentAndCommentMarker` function, to handle single-line comments such that the comment marker is outside of the socket (and therefore, can't be deleted)
* Changed how the `ENTER_KEY` is handled when inside a socket in a comment: Previously, nothing happened. Now, a new single-line comment is created on the next line with the focus in the new socket so that typing can continue.
* Fixed some issues with how handwritten blocks are constructed by properly parenting the `TextToken` and `Socket` that belong to the new block.
* Fixed an issue where `@setCursor` would cause comments to be reparsed - and blank comments would fail to parse
_NOTE: multi-line comments still have issues and this PR does not address those issues_
_NOTE: single-line comment support was added for JavaScript and CoffeeScript only_

### DIFF
--- a/package.json
+++ b/package.json
@@ -12,7 +12,7 @@
     "grunt-cli": "latest",
     "grunt-contrib-coffee": "latest",
     "grunt-contrib-concat": "latest",
-    "grunt-contrib-connect": "1.0.0",
+    "grunt-contrib-connect": "latest",
     "grunt-contrib-cssmin": "latest",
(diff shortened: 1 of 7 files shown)


cpirich:
@Bjvanminnen @joshlory

cpirich:
@Bjvanminnen @joshlory

Bjvanminnen:
lgtm

PR #160, opened 2016-05-13
paramButtons: respect lockZeroParamFunctions and only create buttons when minArgs, maxArgs allow it

cpirich:
* When `lockZeroParamFunctions` is `true`, we don't show the `addButton` or `subtractButton` for function calls that currently have zero parameters
* Selectively show the `addButton` and `subtractButton` in function calls based on `minArgs` and `maxArgs` data that may have been supplied as part of the list of `known` functions. Now, you won't see a `subtractButton` when you are already using the minimum number of arguments, and similarly, you won't see an `addButton` when you are already using the maximum number of arguments.

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -593,14 +593,22 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
       when 'CallExpression', 'NewExpression'
         known = @lookupKnownName node
         blockOpts = {}
+        argCount = node.arguments.length
         if known?.fn
           showButtons = known.fn.minArgs? or known.fn.maxArgs?
+          minArgs = known.fn.minArgs or 0
+          maxArgs = if known.fn.maxArgs? then known.fn.maxArgs else Infinity
         else
-          showButtons = @opts.paramButtonsForUnknownFunctions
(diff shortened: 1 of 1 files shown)


cpirich:
@Bjvanminnen

src/languages/javascript.coffee
```diff
@@ -593,14 +593,22 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
       when 'CallExpression', 'NewExpression'
         known = @lookupKnownName node
         blockOpts = {}
+        argCount = node.arguments.length
         if known?.fn
           showButtons = known.fn.minArgs? or known.fn.maxArgs?
+          minArgs = known.fn.minArgs or 0
+          maxArgs = if known.fn.maxArgs? then known.fn.maxArgs else Infinity
```
Bjvanminnen: I think coffeescript allows us to write this more tersely as `known.fn.maxArgs ? Infinity`

Bjvanminnen:
one nit, otherwise lgtm

src/languages/javascript.coffee
cpirich: Nice, I was trying to figure that out. Will correct in next PR, coming your way shortly..

PR #161, opened 2016-05-13
allow objectDropdown in known.anyobj, allow changing socket in dropdownOnly mode

cpirich:
* Allow a known function that is considered `known.anyobj` (starting with `*.`) to specify an `objectDropdown` - which is essentially the same as a parameter dropdown, but instead applies to the `callee.object` (the left side of the `MemberExpression`)
* A little known feature allows you to specify `dropdownOnly` inside your dropdown object, which makes the socket non-editable. Unfortunately, the dropdown couldn't change the socket text in this case as well, so I fixed that.
* PR feedback from last PR (https://github.com/droplet-editor/droplet/pull/160): use CoffeeScript existential operator

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -2524,6 +2524,9 @@ Editor::showDropdown = (socket = @getCursor(), inPalette = false) ->
       if inPalette
         @populateSocket socket, text
         @redrawPalette()
+      else if not socket.editable()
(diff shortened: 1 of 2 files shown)


cpirich:
@Bjvanminnen

Bjvanminnen:
lgtm

PR #163, opened 2016-06-08
Python mode dropdown and block color

ttsuchiya:
Also added the example html.
TODOs: focus & color bug, keyword vs keyword as value (socketable), parens drag & drop

### DIFF
--- a/example/example-python.coffee
+++ b/example/example-python.coffee
@@ -1,0 +1,207 @@
+# Droplet config editor
+window.expressionContext = {
+  prefix: 'a = '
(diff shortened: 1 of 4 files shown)


ttsuchiya:
@dabbler0 I adapted the old python mode (aka. text-paste) to the most recent master branch, as much as I can figure out.
Remaining issues:
- dropdown menu items don't popup anymore with mouse click
- handle button for if/else is broken
- cannot drop a block from a paren to document
- cannot drop a block to an indent

PR #168, opened 2016-06-22
Allow parsing of bare JavaScript ObjectExpression

joshlory:
Prior to this change, a socket reparse of `{a: 1}` would create a `LabeledStatement` node instead of `ObjectExpression`.

object-expression-in-socket

`LabeledStatement` is an unrecognized node, so `Editor::reparse` doesn't set the correct cursor location.  Subsequent calls to `getCursor`/`setCursor` throw an exception.  To the user it appears the editor is frozen.

This change wraps the program text with parentheses if the text starts with '{'.  Chrome fixed a similar issue in the Dev Tools console (https://bugs.chromium.org/p/chromium/issues/detail?id=499864), however they also check that the text ends with '}'.

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -156,6 +156,9 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
     @lines = @text.split '\n'
 
   markRoot: ->
+    # Socket reparse of `{a: 1}` should be treated as ObjectExpression.
+    @text = "(#{@text})" if @text[0] == '{'
+
     tree = acorn.parse(@text, {
       locations: true
       line: 0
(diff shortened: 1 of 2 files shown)


src/languages/javascript.coffee
```diff
@@ -156,6 +156,8 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
     @lines = @text.split '\n'
 
   markRoot: ->
+    @text = "(#{@text})" if @text[0] == '{'
```
Bjvanminnen: worth a comment explaining why we're doing this here?

joshlory: Amended the previous commit (now ebac9d2db554d3f53f9d582cead768d42f8d8f1b).

Bjvanminnen:
lgtm

joshlory:
I opened this PR against the wrong branch.  See PR https://github.com/droplet-editor/droplet/pull/170.

PR #169, opened 2016-06-23
Cloud9 integration -- Sessions & C support

dabbler0:
- Support for editing sessions bound to underlying Ace sessions, containing value, scroll position, etc.
 - Full support for C mode with a better ANTLR framework.
   - New parser callbacks, which needed to be inserted into the old modes: `stringFixer` (optional), `parseComment` (required).
 - A number of bug fixes:
   - There was a bug where reparsing a socket would sometimes cause the cursor to get lost due the reparse propagating both up and down the tree simultaneously. The reparse now propagates only up the tree, resolving this issue. *However,* this breaks the CoffeeScript quote autoescape feature, so this was reimplemented with a new Parser callback, `stringFixer`.
   - There was a bug with the view `absorbCache()` function, which made blocks grow arbitrarily tall when you lasso-selected them. This is now fixed. 
   - There was a bug where you couldn't click the scrollbar when a block was underneath it. This is now fixed.
   - Multiline comments are now consolidated into one block instead of being listed as a bunch of separate line.
 - There is now a new alternative constructor for `droplet.Editor` that accepts an existing Ace editor to bind to instead of a div to populate.
 - `viewSettings` can now be passed in through the options object if desired.

@joshlory @Bjvanminnen @cpirich
 - This may be an API-breaking change; all functions that hung on `editor` should still work the same, but many properties like `view`, `mode`, and `scrollOffsets` are now in `editor.session` instead of `editor`.
 - This change fixes the bug where the cursor used to sometimes get lost whenever a socket reparse failed, causing the editor to get into an unrecoverable state (e.g. #168). This should no longer happen.
 - This change fixes the bug where the scrollback was unclickable when a block was underneath it.

@ttsuchiya
This change makes a number of modifications to `treewalk.coffee`, some of which are replicates of changes you made for Python mode; you may want to take a look.


antlr/C.g4
```diff
@@ -238,7 +238,7 @@ typeSpecifier
     ;
 
 structOrUnionSpecifier
-    :   structOrUnion Identifier? '{' structDeclarationList '}'
```
davidbau: Question - why the change in the C grammar?  Is this just a new version from antlr, or did you have to rewrite the grammar yourself for some purpose?

src/languages/c.coffee
```diff
@@ -3,42 +3,256 @@
...
+     node.children[0].children[0].type is 'primaryExpression' and
+     node.children[0].children[0].children[0].type is 'Identifier' and
+     node.children[0].children[0].children[0].data.text of opts.knownFunctions
+    return opts.knownFunctions[node.children[0].children[0].children[0].data.text].color
+  else if node.type is 'specialMethodCall' and node.children[0].data.text of opts.knownFunctions
+    return opts.knownFunctions[node.children[0].data.text].color
+  return null
+
+config.SHAPE_CALLBACK = (opts, node) ->
+  return null unless opts.knwonFunctions?
```
davidbau: typo: knownFunctions?  (Is there not a test that exercises this?)

src/languages/c.coffee
```diff
@@ -3,42 +3,256 @@
...
+  ['expression', 'value'],
+  ['parameterDeclaration', 'command'],
+  ['unaryExpression', 'value'],
+  ['typeName', 'value'],
+  ['initializer', 'value'],
+  ['castExpression', 'value'],
+]
+
+SHAPE_RULES = [
+  ['blockItem', 'block-only'],
```
davidbau: Next to each of these add a comment with an example of the syntax represented by the node.

src/languages/c.coffee
```diff
@@ -3,42 +3,256 @@
...
+  }
+}
+
+config.SHOULD_SOCKET = (opts, node) ->
+  unless opts.knownFunctions? and ((node.parent? and node.parent.parent? and node.parent.parent.parent?) or
+      node.parent?.type is 'specialMethodCall')
+    return true
+
+  # If it is a function call, and we are the first child
+  if (node.parent.type is 'primaryExpression' and
```
davidbau: This is a very long set of conditions.  It deserves a comment showing an example expression or tree to justify what is being tested.

src/languages/c.coffee
```diff
@@ -3,42 +3,256 @@
...
+     node.parent.parent is node.parent.parent.parent.children[0] or
+     node.parent.type is 'specialMethodCall') and
+     node.data.text of opts.knownFunctions
+    return false
+  return true
+
+config.COLOR_CALLBACK = (opts, node) ->
+  return null unless opts.knownFunctions?
+
+  if node.type is 'postfixExpression' and
```
davidbau: This has some similarities to the expression just above.  Also needs to be explained. And can anything be factored into a common test function?

src/languages/c.coffee
```diff
@@ -3,42 +3,256 @@
...
+    return opts.knownFunctions[node.children[0].children[0].children[0].data.text].color
+  else if node.type is 'specialMethodCall' and node.children[0].data.text of opts.knownFunctions
+    return opts.knownFunctions[node.children[0].data.text].color
+  return null
+
+config.SHAPE_CALLBACK = (opts, node) ->
+  return null unless opts.knwonFunctions?
+
+  if node.type is 'postfixExpression' and
+     node.children.length in [3, 4] and
```
davidbau: Also needs to be explained.

src/model.coffee
```diff
@@ -1017,14 +1021,18 @@ exports.SocketStartToken = class SocketStartToken extends StartToken
...
+    if opts.preserveEmpty and @prev is @container.start or
         @prev.type is 'text' and @prev.value is ''
       return @container.emptyString
+
+    # Otherwise, do nothing, and allow the socket to stringify to ''
     else ''
 
 exports.Socket = class Socket extends Container
-  constructor: (@emptyString, @precedence = 0, @handwritten = false, @classes = [], @dropdown = null) ->
+  constructor: (@emptyString, @precedence = 0, @handwritten = false, @classes = [], @dropdown = null, @parseContext = null) ->
```
davidbau: Here you're adding a new field to Socket and it strikes me that they all need to be documented.
Can you break this out to one field per line and add a comment explaining the purpose of each one, including the new one parseContext?

src/parser.coffee
```diff
@@ -101,6 +104,14 @@ exports.Parser = class Parser
 
     @addMarkup block, opts.bounds, opts.depth
 
+  # flagToRemove
```
davidbau: What is flagToRemove for?  Add a comment.

src/parser.coffee
```diff
@@ -302,18 +386,26 @@ exports.Parser = class Parser
           lastIndex = indentDepth
 
         for mark in markupOnLines[i]
+          # If we are not supposed to be flagging to remove,
```
davidbau: I don't understand the flagToRemove mechanism.  The comment here (or elsewhere) needs to explain the situation a bit more clearly.

src/parser.coffee
```diff
@@ -204,35 +216,72 @@ exports.Parser = class Parser
   # Construct a handwritten block with the given
   # text inside
```
davidbau: Add a comment here describing the implementation strategy in a few words.

src/parser.coffee
```diff
@@ -257,6 +306,8 @@ exports.Parser = class Parser
     stack = []
     document = new model.Document(); head = document.start
 
+    currentlyCommented = false
```
davidbau: applyMarkup is a pretty long method now.  At the top of it, add a comment describing (1) what the job of the method is, (2) what the hard things are that need to be tracked and (3) a brief overview of the implementation strategy.

davidbau:
General question: I see all the session-sensitive tests (great) but I don't see much code in the diffs setting up the new session objects.  Are those not in the pull request because they were merged previously, or are there any missing files?

src/helper.coffee
```diff
@@ -179,3 +184,68 @@ exports.deepEquals = deepEquals = (a, b) ->
...
+             replace(/((?:^|[^\\])(?:\\\\)*)\\"/g, '$1"').
+      replace(/'/g, "\\'") + quotechar
+  return result
+
+# A naive dictionary mapping arbitrary objects to arbitrary objects, for use in
+# ace-to-droplet session matching.
+#
+# May replace with something more sophisticated if performance becomes an issue,
+# but we don't envision any use cases where sessions flip really fast, so this is unexpected.
+exports.PairDict = class PairDict
```
davidbau: I don't see this class being used anywhere.

dabbler0: It is used in the Sessions code. I think that code must have already been merged, since it's not showing up in the diff.

PR #170, opened 2016-06-27
Allow parsing of bare JavaScript ObjectExpression

joshlory:
Re-opening PR https://github.com/droplet-editor/droplet/pull/168 against the correct branch.

---

Prior to this change, a socket reparse of `{a: 1}` would create a `LabeledStatement` node instead of `ObjectExpression`.

object-expression-in-socket

`LabeledStatement` is an unrecognized node, so `Editor::reparse` doesn't set the correct cursor location.  Subsequent calls to `getCursor`/`setCursor` throw an exception.  To the user it appears the editor is frozen.

This change wraps the program text with parentheses if the text starts with '{'.  Chrome fixed a similar issue in the Dev Tools console (https://bugs.chromium.org/p/chromium/issues/detail?id=499864), however they also check that the text ends with '}'.

### DIFF
--- a/src/languages/javascript.coffee
+++ b/src/languages/javascript.coffee
@@ -156,6 +156,9 @@ exports.JavaScriptParser = class JavaScriptParser extends parser.Parser
     @lines = @text.split '\n'
 
   markRoot: ->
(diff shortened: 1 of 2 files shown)


Bjvanminnen:
lgtm

PR #178, opened 2016-07-11
C if else only

dabbler0:
Like #176 but without the new plumbing for annotations required for function and declarator buttons.


dabbler0:
Abandoned in favor of #181

PR #179, opened 2016-07-11
Arbitrary buttons

dabbler0:
Shifts from having just +/- buttons to allowing any number of buttons with arbitrary glyphs. Also adds some plumbing in ANTLR to make more +/- buttons possible in C mode, and implements +/- buttons for several constructs in C mode.


dabbler0:
Abandoned in favor of #181

PR #181, opened 2016-07-13
Container buttons

dabbler0:
Introduces a new Model entity, `buttonContainer`, which is invisible in the view except that it has buttons at the end of it. The advantage of requiring a container is that it:
 - defines a scope for the button mutation
 - provides reparse boundaries, so that the entire block context doesn't need to always be reparsed, and
 - it usually has semantic meaning, unlike the old arbitrary-token-in-model approach.


dabbler0:
Closing to merge directly into master

PR #187, opened 2016-07-20
Python mode (aka text-paste) branch

ttsuchiya:


src/treewalk.coffee
```diff
@@ -112,6 +118,7 @@ exports.createTreewalkParser = (parse, config, root) ->
               color: @getColor node, rules
               classes: padRules(wrapRules ? rules).concat(@getShape(node, rules))
               parseContext: rules[0] #(if wrap? then wrap.type else rules[0])
+              buttons: config.BUTTON_CALLBACK(@opts, node) ? null
```
dabbler0: This breaks C mode right now. Can you either add an empty `BUTTON_CALLBACK` to C mode, or make `BUTTON_CALLBACK` optional by checking to see whether it exists before calling it?

Along with fixing the indent thing mentioned below, this should theoretically make the tests pass again.

src/languages/python.coffee
```diff
@@ -36,52 +42,156 @@ getBounds = (node, lines) ->
...
+  if getArgNum(node) is null
+    return opts.functions?[getFunctionName(node)]?.color ? null
+
+insertButton = (opts, node) ->
+  if node.type is 'if_stmt'
+    return {addButton: true}
+  else
+    return null
+
+handelButton = (text, button, oldBlocks) ->
```
dabbler0: Typo?

src/treewalk.coffee
```diff
@@ -175,7 +182,9 @@ exports.createTreewalkParser = (parse, config, root) ->
                 break
               else unless i is 0
                 end = child.bounds.start
-                if @lines[end.line][...end.column].trim().length is 0
+
+                # @lines[end.line] was breaking in python mode (index out of range)
+                if @lines[end.line-1][...end.column].trim().length is 0
```
dabbler0: C mode needs it to be `@lines[end.line]`. The purpose is for cases like

```
if (a == b) {
  //whatever
}
```

The indent ends on the closing brace `}`, but we want it to end at the end of the previous line (`//whatever`), so we move it back by one.

Instead of changing to `@lines[end.line - 1]`, could you do a check like `if end.line < @lines and @lines[end.line`...etc? What are some program examples of where this breaks in python mode?

dabbler0:
Looks good aside from comments above.

PR #194, opened 2017-04-07
Include `sourcesContent` in the generated source map

joshlory:
Needed for New Relic un-minification.


Bjvanminnen:
lgtm. should i rebuild with this before merging latest droplet in https://github.com/code-dot-org/code-dot-org/pull/14265?

joshlory:
No worries if we don't merge the updated source map with your hotfix.  I'm still manually uploading source map to New Relic at this point.

PR #195, opened 2017-04-14
Don't throw if a reparsed list has no parent

joshlory:
Fixes the following edit:

`f(1);`
  -> Edits `1` to `); g(`
  -> Reparse FAILS
  -> Fall back to reparsing the parent
  -> Reparse `f(); g();` as two separate blocks
  -> Finished

Sample edits that throw an exception prior to this change:

screen shot 2017-04-12 at 2 28 20 pm

screen shot 2017-04-12 at 2 21 38 pm

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -2260,7 +2260,7 @@ Editor::reparse = (list, recovery, updates = [], originalTrigger = list) ->
 
   try
     newList = @mode.parse list.stringifyInPlace(),{
-      wrapAtRoot: parent.type isnt 'socket'
+      wrapAtRoot: parent?.type isnt 'socket'
       context: context
     }
   catch e


Bjvanminnen (approved):

joshlory:
To understand why this happens, set a conditional breakpoint in `Editor::reparse` where `parent` is assigned:

screen shot 2017-04-14 at 4 54 35 pm

Editing a socket in a way that breaks the parent block into two or more blocks hits this breakpoint:

screen shot 2017-04-14 at 4 56 41 pm

At this point `parent.type` throws because `parent` is null:

screen shot 2017-04-14 at 4 58 08 pm

The `catch` block suppresses the exception, but I believe the intent of this try/catch is to catch parser errors, not the error thrown when `parent` is null.

PR #198, opened 2017-04-17
Pin packages to specific versions

Bjvanminnen:
Following up on #197 this pins all package versions. This may be somewhat more contentious, but it means we don't have to worry about things breaking because of updated packages. Instead, updating dependencies to newer versions becomes an explicit steps where you update the version in package.json.

I got these versions by looking at the current versions in my node_modules.


joshlory (approved):
LGTM 👍  thanks for tracking these down & fixing!

PR #200, opened 2017-06-26
implement subtract button on else clauses

cpirich:
* Subtract button added to else clauses in JavaScript
  * When pressed, the button removes the final `else` or `else if` clause and the statements within. (Undo will bring back anything that was deleted). If we find that we don't like this behavior, we can modify the button handling to try to extract only the `} else {` or `} else if (__) {`, effectively moving the code within the final clause up into the previous clause.

subtractelseclauses


davidsbailey (approved):
Looks good!

davidsbailey:
I'm going to go ahead and merge this for you so that I can prepare to pull it into the code-dot-org repo.

PR #201, opened 2017-06-27
fix svg dropdowns

davidsbailey:
Prior to this change there was the following bug with dropdowns in svg droplet:
droplet-dropdown-issues

This only happened when the cursor was previously in a different socket within the same block as the dropdown that was being opened.

The problem was that the `showDropdown` method calls `setTimeout` to wait for a render before showing the dropdown, but it was keeping a reference to the old model id for the dropdown, which had become invalid. The solution is to refresh the socket from the cursor position after the render, before opening the dropdown.


cpirich (approved):
Nice work!

PR #202, opened 2017-07-13
replace parentElement with parentNode to fix IE

davidsbailey:
parentElement is not defined on SVG elements in Internet Explorer, but parentNode is. This brings the time taken to redraw the block palette down from 7 seconds to 0.05 seconds in IE.


Bjvanminnen (approved):
Nice find. LGTM

Bjvanminnen:
@dabbler0 @davidbau This is a fix you guys are probably interested in pulling into master as well.

PR #203, opened 2017-07-19
dont clobber other classnames on .droplet-hover-div

davidsbailey:
We need to preserve the `.tooltipstered` class on the `.droplet-hover-div` when the palette is redrawn, so that we can hide the tooltip here https://github.com/code-dot-org/code-dot-org/blob/staging/apps/src/blockTooltips/DropletBlockTooltipManager.js#L117-L121 when a block is selected.

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -830,7 +830,7 @@ Editor::redrawPalette = ->
     # enable tooltips, making sure not to overwrite any additional classes
     # that may have been added (e.g. by tooltipster)
     paletteBlockClass = paletteBlockView.group.element.getAttribute('class')
-    unless paletteBlockClass && paletteBlockClass.includes('droplet-hover-div')
+    unless paletteBlockClass && paletteBlockClass.indexOf('droplet-hover-div') > -1
       paletteBlockView.group.element.setAttribute('class', 'droplet-hover-div')
     paletteBlockView.group.element.setAttribute 'title', entry.title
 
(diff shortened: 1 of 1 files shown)


src/controller.coffee
```diff
@@ -827,8 +827,11 @@ Editor::redrawPalette = ->
 
     paletteBlockView.group.element.setAttribute 'data-id', entry.id
 
-    # enable tooltips
-    paletteBlockView.group.element.setAttribute('class', 'droplet-hover-div')
+    # enable tooltips, making sure not to overwrite any additional classes
+    # that may have been added (e.g. by tooltipster)
+    paletteBlockClass = paletteBlockView.group.element.getAttribute('class')
+    unless paletteBlockClass && paletteBlockClass.includes('droplet-hover-div')
+      paletteBlockView.group.element.setAttribute('class', 'droplet-hover-div')
```
Bjvanminnen: If the element has just the class name from tooltipster, and does not have 'droplet-hover-div', won't we still blow away the other class?

Should the last line instead be ` paletteBlockView.group.element.setAttribute('class', paletteBlockClass + ' droplet-hover-div')`

davidsbailey: In our case, we only tooltipster this class by finding first via the `.droplet-hover-div` class, so there's no actual danger of clobbering it. If you think this reads poorly or sets us up badly for the future, I'm happy to make this change though.

Bjvanminnen: I think my main concern is that right now this feels like a code.org specific issue (presumably we're the ones using tooltipster to add classes). As much as possible, a fix should address the general case.

It may be the case that we don't actually have to worry about anyone else setting this class name, but it feels cleaner/less destructive to me to have code that says "add this class if we dont already have it" rather than "replace all the classes with this class if we dont have this class"

davidsbailey: Makes sense. done

Bjvanminnen (approved):

PR #204, opened 2017-08-08
Redraw after toggle palette

davidsbailey:
This fixes the issue where the droplet block text disappears after the toolbox is shown or hidden. I'm not adding a test for this, since this regression was already caught by our existing eyes tests.


davidsbailey:
before:
toolbox-toggle-before

after:
toolbox-toggle-after

Bjvanminnen (approved):

PR #208, opened 2019-07-09
Add nvmrc to droplet and fix failing unit tests

jmkulwik:
This enforces node v8.15.0
Additionally, this fixes two unit tests that were failing due to commit 07865f25a1f0af430f12e217349d96b8a80190a1


test/src/tests.coffee
```diff
@@ -796,7 +796,7 @@ asyncTest 'Controller: showPaletteInTextMode false', ->
     strictEqual aceEditor.style.left, '0px'
     editor.performFreezeAnimation 10, 10, ->
       strictEqual paletteWrapper.style.left, '0px'
-      strictEqual aceEditor.style.left, '-9999px'
+      strictEqual aceEditor.style.left, '0px'
```
davidsbailey, line 799: `-9999px` is usually used to hide something off the screen. @cpirich are you confident that we are not accidentally showing something that should be hidden after commit 07865f2 ?

davidsbailey (approved):

davidsbailey:
I would suggest also adding a section like this to package.json: https://github.com/code-dot-org/code-dot-org/blob/bf0500cb5c55c8df6c08f502d0a9aa31c805d36a/apps/package.json#L6-L9

cpirich (approved):

PR #209, opened 2019-07-09
Update dropdowns to extend outside the coding workspace

jmkulwik:
Relates to this PR: https://github.com/code-dot-org/code-dot-org/pull/29588
Ticket: https://codedotorg.atlassian.net/browse/STAR-461

Droplet dropdowns now render in front of other parts of the screen and respect their z-index.

In the past, the code workspace had all overflows set to "hidden" this effectively caused the droplet dropdowns to ignore their z-index and hide any section that interacted with another component on the screen. It would appear to render under these components. This "hidden" overflow was used to make the droplet palette "disappear" when the palette was closed because it would interact with elements outside the code workspace. This change removes the "hidden" overflow. Additionally, it adjusts the palette to disappear under the code workspace when closed rather than disappear under the app screen when closed.

**Dropdown before:**
image

**Dropdown after:**
image

**Palette before:**
oldToolbox

**Palette after:**
newToolbox2

**[CS50 Compatibility] After these changes**
cs50Demo

**FOR RECORDS (NO LONGER ACCURATE - fixed after comments) Palette after:**
newToolbox

### DIFF
--- a/css/droplet.css
+++ b/css/droplet.css
@@ -71,7 +71,7 @@
   position: absolute;
   top: 0; bottom: 0; width: 300px;
   max-width: 40%;
-  z-index:0;
+  z-index: 0;
   -webkit-user-select: none;
   -khtml-user-select: none;
   -moz-user-select: none;
--- a/css/droplet.css
+++ b/css/droplet.css
@@ -71,7 +71,7 @@
   position: absolute;
   top: 0; bottom: 0; width: 300px;
   max-width: 40%;
-  z-index: 257;
+  z-index:0;
   -webkit-user-select: none;
   -khtml-user-select: none;
   -moz-user-select: none;
@@ -263,7 +263,7 @@
   max-height: 150px;
   position: absolute;
   background-color: #FFF;
-  z-index: 10000;
+  z-index: 500;
   border: 1px solid #DDD;
   border-radius: 2px;
   overflow-y: auto;
--- a/test/src/tests.coffee
+++ b/test/src/tests.coffee
@@ -792,7 +792,7 @@ asyncTest 'Controller: showPaletteInTextMode false', ->
     states.push usingBlocks
 
   editor.performMeltAnimation 10, 10, ->
-    strictEqual paletteWrapper.style.left, '-270px'
+    strictEqual paletteWrapper.style.left, '0px'
     strictEqual aceEditor.style.left, '0px'
     editor.performFreezeAnimation 10, 10, ->
       strictEqual paletteWrapper.style.left, '0px'
@@ -840,7 +840,7 @@ asyncTest 'Controller: enablePalette false', ->
   strictEqual dropletWrapper.style.left, '270px'
 
   verifyPaletteHidden = ->
-    strictEqual paletteWrapper.style.left, '-270px'
+    strictEqual paletteWrapper.style.left, '0px'
     strictEqual dropletWrapper.style.left, '0px'
     start()
 
--- a/test/src/uitest.coffee
+++ b/test/src/uitest.coffee
@@ -976,9 +976,16 @@ asyncTest 'Controller: Session switch test', ->
       equal editor.paletteHeader.childElementCount, 0, 'Palette header now empty'
 
       equal editor.paletteWrapper.style.left is '0px', true, 'Using blocks'
+      equal editor.dropletElement.style.top is '0px', true, 'droplet visible'
+      equal editor.dropletElement.style.left is '270px', true, 'droplet next to palette'
+      equal editor.aceElement.style.top is '-9999px', true, 'ace element offscreen'
+
       editor.setEditorState(false)
 
       equal editor.paletteWrapper.style.left is '0px', true, 'No longer using blocks'
+      equal editor.dropletElement.style.top is '-9999px', true, 'droplet offscreen'
+      equal editor.aceElement.style.top is '0px', true, 'ace element visible'
+      equal editor.aceElement.style.left is '0px', true, 'palette covered by ace element'
     ),
     (->
       editor.aceEditor.setSession originalSession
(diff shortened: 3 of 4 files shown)


test/src/tests.coffee
```diff
@@ -821,7 +821,7 @@ asyncTest 'Controller: showPaletteInTextMode true', ->
     strictEqual aceEditor.style.left, '270px'
     editor.performFreezeAnimation 10, 10, ->
       strictEqual paletteWrapper.style.left, '0px'
-      strictEqual aceEditor.style.left, '-9999px'
+      strictEqual aceEditor.style.left, '270px'
```
jmkulwik: Part of PR https://github.com/droplet-editor/droplet/pull/208

test/src/tests.coffee
```diff
@@ -792,11 +792,11 @@ asyncTest 'Controller: showPaletteInTextMode false', ->
     states.push usingBlocks
 
   editor.performMeltAnimation 10, 10, ->
-    strictEqual paletteWrapper.style.left, '-270px'
+    strictEqual paletteWrapper.style.left, '0px'
     strictEqual aceEditor.style.left, '0px'
     editor.performFreezeAnimation 10, 10, ->
       strictEqual paletteWrapper.style.left, '0px'
-      strictEqual aceEditor.style.left, '-9999px'
+      strictEqual aceEditor.style.left, '0px'
```
jmkulwik: Part of PR https://github.com/droplet-editor/droplet/pull/208

.nvmrc
```diff
@@ -0,0 +1 @@
+8.15.0
```
jmkulwik: Part of PR https://github.com/droplet-editor/droplet/pull/208

css/droplet.css
```diff
@@ -263,7 +262,7 @@
   max-height: 150px;
   position: absolute;
   background-color: #FFF;
-  z-index: 10000;
+  z-index: 500;
```
jmkulwik, line 266: This matches our enforced value from https://github.com/code-dot-org/code-dot-org/blob/1e3ae0c714352eb2341c0b5b548009bc98c74283/apps/style/applab/style.scss#L489
z-index of 10000 was causing problems as it was showing up in front of dialogs. We forcibly set the z-index to 500 with !important. Setting to 500 will make this more usable for other users of the droplet repo as well.

davidsbailey:
👍 The parts relating to node versions look good to me.

cpirich:
I worry that these changes may make it difficult / unlikely to merge our branch back into `master`. The reason for my concern is that there are other clients that use the droplet editor (at least, [pencilcode](https://pencilcode.net/), maybe others). Without testing the impact on these editors, it's hard to say what the impact will be.

css/droplet.css
```diff
@@ -71,7 +71,7 @@
   position: absolute;
   top: 0; bottom: 0; width: 300px;
   max-width: 40%;
-  z-index: 257;
+  z-index:0;
```
joshlory: Tiny style nit: missing space.
```suggestion
  z-index: 0;
```

test/src/uitest.coffee
```diff
@@ -978,7 +978,7 @@ asyncTest 'Controller: Session switch test', ->
       equal editor.paletteWrapper.style.left is '0px', true, 'Using blocks'
       editor.setEditorState(false)
 
-      equal editor.paletteWrapper.style.left is '0px', false, 'No longer using blocks'
+      equal editor.paletteWrapper.style.left is '0px', true, 'No longer using blocks'
```
joshlory, line 985: Should `editor.setEditorState(false)` have modified the paletteWrapper metrics?

joshlory: Ah I see, because it slides of the screen to the right instead of the left. Should we update the test to check the `.left` of the main block area?

joshlory:
@cpirich what's the best way to get the behavior we want (drop downs not clipped) while still making an upstream merge possible? I can think of:
 - Reparent the dropdown element under the document body instead of the editor element.
 - Update the display logic so it bumps away from the edge and never goes offscreen.
 - More testing with other apps that use the Droplet Editor.

cpirich:
> @cpirich what's the best way to get the behavior we want (drop downs not clipped) while still making an upstream merge possible? I can think of:
> 
> * Reparent the dropdown element under the document body instead of the editor element.

This was the idea I proposed, but I don't know if it is easy to do. At least it is constrained to dropdowns, though it does require that our `codeWorkspace` lose its `overflow: hidden` which *may* have some impact when it is hosting Blockly or Bramble.

> * Update the display logic so it bumps away from the edge and never goes offscreen.

I think we already did some of this (go up if near the bottom, go down if near the top). It seems really familiar (I think I made ace autocomplete do something similar). But it may not be enough. The screenshots above are dealing with a case where the window has been resized so small that there is now way to fit the dropdown within the window dimensions. Though I guess we could make the dropdown just show 1-2 items in height in that case. I certainly don't mind this approach, as this is what we've tried to do in the past.

> * More testing with other apps that use the Droplet Editor.

This is a reasonable approach if we plan to make a whole bunch of changes in Droplet and/or want to take over ownership of the repo. But if we're just making a surgical change or two, it is probably more effort than we'd want to take on.

jmkulwik:
> > @cpirich what's the best way to get the behavior we want (drop downs not clipped) while still making an upstream merge possible? I can think of: 
> > ...
> > * More testing with other apps that use the Droplet Editor.
> 
> This is a reasonable approach if we plan to make a whole bunch of changes in Droplet and/or want to take over ownership of the repo. But if we're just making a surgical change or two, it is probably more effort than we'd want to take on.

This time around, I chose to do this method. Not because we plan/want to take ownership of the repo, but to see if testing in this manner is a future possibility. While there is additional work involved, this may be a viable possibility in the future. It took a long time to setup, but wasn't to difficult once the setup was finished. Most of the difficulty came from understanding the state of the CS50 usage of droplet. 

The results of the testing can be seen in the "CS50 Compatibility" gif above. The is no regression on the CS50 project. Additionally, there is some value added since the dropdowns extend slightly further - to the outside edge (but not beyond) of the coding workspace.

test/src/uitest.coffee
jmkulwik: Fixed the tests.

joshlory (approved):
Thanks for testing on the CS50 editor as well!

+cc @kzidane

cpirich (approved):

PR #211, opened 2019-07-23
Remove reliance on calculated width and prevent workspace from displaying outside its container

jmkulwik:
Relates to this PR in the code-dot-org repo: https://github.com/code-dot-org/code-dot-org/pull/29849

**OLD:**
workspaceWithoutDropletChanges

**NEW:**
workspaceWithDropletChanges

**CS50 workspace after changes**
cs50withChanges

### DIFF
--- a/src/controller.coffee
+++ b/src/controller.coffee
@@ -3646,8 +3646,9 @@ Editor::performMeltAnimation = (fadeTime = 500, translateTime = 1000, cb = ->) -
       @dropletElement.style.transition = ''
 
       # Translate the ACE editor div into frame.
-      @aceElement.style.top = '0px'
-      @aceElement.style.bottom = '0px'
+      @aceElement.style.top =
+        @aceElement.style.right =
+        @aceElement.style.bottom = '0px'
       if @session.showPaletteInTextMode and @session.paletteEnabled
         @aceElement.style.left = "#{@paletteWrapper.clientWidth}px"
       else
@@ -3657,9 +3658,8 @@ Editor::performMeltAnimation = (fadeTime = 500, translateTime = 1000, cb = ->) -
       #  @paletteWrapper.style.top = '-9999px'
       #  @paletteWrapper.style.left = '-9999px'
 
-      @dropletElement.style.top = '-9999px'
-      @dropletElement.style.bottom = '9999px'
-      @dropletElement.style.left = '-9999px'
+      @dropletElement.style.top = @dropletElement.style.left = '-9999px'
+      @dropletElement.style.bottom = @dropletElement.style.right = '9999px'
 
       # Finalize a bunch of animations
       # that should be complete by now,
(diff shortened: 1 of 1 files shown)


src/controller.coffee
```diff
@@ -3651,6 +3658,7 @@ Editor::performMeltAnimation = (fadeTime = 500, translateTime = 1000, cb = ->) -
       #  @paletteWrapper.style.left = '-9999px'
 
       @dropletElement.style.top = '-9999px'
+      @dropletElement.style.bottom = '9999px'
```
cpirich: should you also specify `right` here for symmetry?

src/controller.coffee
```diff
@@ -3703,13 +3711,14 @@ Editor::performFreezeAnimation = (fadeTime = 500, translateTime = 500, cb = ->)-
...
       @showScrollbars false
-      @dropletElement.style.width = @wrapperElement.clientWidth + 'px'
+      @dropletElement.style.right = "0px"
 
       @redrawMain noText: true
 
       @currentlyAnimating_suppressRedraw = true
 
       @aceElement.style.top = "-9999px"
+      @aceElement.style.bottom = "9999px"
```
cpirich: same as above, should you specify `right` as well?

src/controller.coffee
```diff
@@ -3861,13 +3872,15 @@ Editor::enablePalette = (enabled) ->
         #@paletteWrapper.style.left = '-9999px'
 
         @currentlyAnimating = false
-
-        @redrawMain()
+        @resize()
+        @resizeNubby()
+        @paletteWrapper.style.visibility = "hidden"
```
cpirich: I'm not totally sure I follow this particular change.

src/controller.coffee
```diff
@@ -3948,9 +3961,8 @@ hook 'populate', 2, ->
     @session.viewports.palette.x = @paletteScroller.scrollLeft
 
 Editor::resizeMainScroller = ->
-  @mainScroller.style.width = "#{@dropletElement.clientWidth}px"
-  @mainScroller.style.height = "#{@dropletElement.clientHeight}px"
```
cpirich, line 3952: was this line never needed? or is just not needed now that you are specifying `bottom` and `top` in other places?

cpirich (commented):
I really like the direction of these changes (it is definitely a better, more CSS friendly way to handle sizing).

I'm a little concerned about the continued reliance on moving things "offscreen" with the large 9999px or -9999px offsets. (but that's not a new problem - just seems to appear in more places now)

And I'm a little concerned at how many different places needed to have tiny changes applied, so it isn't obvious if you've missed some. In the future, we could probably make this cleaner.

src/controller.coffee
jmkulwik: top, bottom, left, and right were already set in droplet.css. This line may have been needed at some point, but isn't anymore.

src/controller.coffee
jmkulwik: (Assuming you are talking about the "visibility" change)

Good point. This was left over from an earlier attempt. Thanks for catching it. I'll remove this and the other "visibility" line.

cpirich (approved):
Looks good! I'm hopeful that this all works, but expect that you'll need a fair amount of manual testing (if you haven't already).

PR #217, opened 2020-11-30
Make UpdateExpression arithmetic (purple) and any-drop

ajpal:
See https://github.com/code-dot-org/code-dot-org/pull/38016


ajpal:
Merging with failing build since it appears the build is broken already (see https://travis-ci.org/github/droplet-editor/droplet/builds/747739230), and I've manually tested in code-studio and it looks good

PR #224, opened 2021-06-09
Updates to fix builds in travis

aaronwaggener:
- Bump node to v8.15.0 in travis-ci
- Fix a naming collision with the `LANGUAGES` environment variable that is set on the travis-ci docker images
- Bump Qunit to v1.23.1 and `grunt-contrib-qunit` to v3.1.0 (it is expected for these versions not to match)
- Fix c++ test

### DIFF
--- a/Gruntfile.coffee
+++ b/Gruntfile.coffee
@@ -7,8 +7,8 @@ livereload = require 'tiny-lr'
 path = require 'path'
 fs = require 'fs'
 
-if process.env.LANGUAGE
-  ignoredLanguages = ["./src/languages/!(#{process.env.LANGUAGE}).coffee"]
+if process.env.DROPLET_LANGUAGE
+  ignoredLanguages = ["./src/languages/!(#{process.env.DROPLET_LANGUAGE}).coffee"]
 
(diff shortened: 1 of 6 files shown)


Gruntfile.coffee
```diff
@@ -7,7 +7,7 @@ livereload = require 'tiny-lr'
 path = require 'path'
 fs = require 'fs'
 
-if process.env.LANGUAGE
+if process.env.DROPLET_LANGUAGE
```
maddiedierker, line 10: do you know what this environment variable is used for? it doesn't seem likely, but i'm wondering if there are any subsequent changes we'll need to make

aaronwaggener: This was the only reference. `ignoredLanguages` is used at https://github.com/droplet-editor/droplet/blob/6251c62190970bebc1ba6b49bd5ccd33123cff2d/Gruntfile.coffee#L81 and with the naming collision was producing incomplete builds that resulted in our tests not being able to run.

maddiedierker (approved):
great find! what a wild naming collision 😄

PR #226, opened 2022-04-16
Support for createSlotForKnownBlock option

breville:
See https://github.com/code-dot-org/code-dot-org/pull/45868 for full details.

Note that `DROPLET_LANGUAGE=javascript grunt dist` was used to build the files in `dist/` that were then copied into the main `code-dot-org` repo.


breville:
Note that it would have been convenient to simply change [this line](https://github.com/droplet-editor/droplet/blob/12225aab74855fd784c49210dd19894b192991cc/src/languages/javascript.coffee#L192) from
```
 return name: full, anyobj: false, fn: @opts.functions[full]
```
to
```
 return name: full, anyobj: @opts.createSocketForKnownBlock, fn: @opts.functions[full]
```
but we wanted to behave slightly differently in our responses, specifically using a precedence of `3` in each case.

PR #227, opened 2022-08-08
Fix JS error

breville:
The New Relic JavaScript error tracking for https://studio.code.org reports its top error as:
```
Error: Cannot read properties of undefined (reading 'unmark')
```

I've not been able to repro, but believe this fix should fix the error, hopefully without any noticeable side effects.


epeach (approved):
:+1:

